PR #5, opened 2021-01-11
Ks 2021 01 add user dashboard skeleton

Rineee:
The 'styling' is basically copied from user_profile.scss 

I will do now: 
- fix broken logout
- only show moderator tab when you are a moderator
- get list of organisations you are part of/following


phillimorland (approved):
awesome!

PR #7, opened 2021-01-11
apps/users: only show public and semi-public projects in profile view

fuzzylogic2000:
It currently also shows the private projects to unauthenticated users.


phillimorland (approved):
cool!

PR #21, opened 2021-01-12
Ks 2020 01 show moderator tab only for moderators

Rineee:


phillimorland (changes requested):
as a normal user, I can still access the /moderation url but there is no tab

phillimorland (approved):
super cool!

PR #22, opened 2021-01-18
Pm 2021 01 activities

phillimorland:
@Rineee is that how you would add it to the view or am I being over complicated?


fuzzylogic2000 (approved):
Very nice! Some styling is still a bit off, but I think that's for next time!?

PR #42, opened 2021-01-21
apps/userdashboard: mobile styling for userdashboard and child pages

phillimorland:


apps/userdashboard/views.py
```diff
@@ -40,6 +40,13 @@ def actions(self):
             actor=self.request.user,
         ).exclude_updates()
 
+    @property
+    def projects_carousel(self):
+        sorted_active_projects, sorted_future_projects, sorted_past_projects =\
+            self.request.user.get_projects_follow_list()
```
fuzzylogic2000, line 46: I guess, we do filter out the private projects here? In apps.users.models? But we need them here. Can we just filter them out for the public profile then @Rineee ?

fuzzylogic2000 (commented):
It looks good in the profile!
But it moved the chevrons (are they called that?) in the dashboard to a weird place:
Screenshot from 2021-01-22 09-47-14

phillimorland:
@fuzzylogic2000 can't seem to resolve the chevron comment but thats now fixed

apps/userdashboard/views.py
Rineee: Yes, thats true, we filter out the private projects in user.get_projects_follow_list(). I will add a parameter to it, so we can only filter them out if we want to.. but lets merge it here like this and I'll add a task?

fuzzylogic2000 (approved):
Now it really looks fine. :thinking:

PR #45, opened 2021-01-25
issues fixes 

### LINKED ISSUE #39: line break in user tile
**URL:** 
**user:** admin
**expected behaviour:** Name stay the same
**behaviour:** name breaks within the word
**important screensize:** desktop and mobile
**device & browser:** firefox 84.02 on macOS Big Sur
**Comment/Question:** Here the line is breaking, is it possible to have longer names displayed in 2 rows? Thanks!

Bildschirmfoto 2021-01-20 um 17 29 36

phillimorland:


adhocracy-plus/templates/includes/userdashboard_mobile_toggle.html
```diff
@@ -17,10 +17,13 @@
                  href="{% url 'userdashboard-overview' %}">
                   {% trans 'Overview' %}
               </a>
+              {% has_perm 'a4_candy_userdashboard.view_moderation_dashboard' user as user_is_moderator %}
```
Rineee, line 20: This throws an error for me, I think you need to load the rules package..

Rineee (approved):
Yeah, coool!

The list of organisations in the following and activities view is not there, but was like this before, so will merge!

PR #51, opened 2021-01-25
apps/user: add exclude_private_projects parameter to get_projects_fol…

Rineee:
…low_list

In project carousel in userdashboard overview, only running and future modules are shown (as in user profile), past projects can be viewed by clicking on 'view all'. Not sure if this is what we want, couldnt really tell from the story description?

The sorting is still a thing, especially when viewing all projects, they are not sorted at all.. left it like this for now, because we have to  find an overall solution.


phillimorland (approved):
cool!

PR #64, opened 2021-01-26
apps/userdashboard/views: add UserDashboardBaseMixin

Rineee:
I moved all the duplicate stuff into a mixin.. but thought its better to leave in views than in a mixin file, what do you think?

### DIFF
--- a/apps/userdashboard/views.py
+++ b/apps/userdashboard/views.py
@@ -8,15 +8,11 @@ from apps.organisations.models import Organisation
 from apps.users.models import User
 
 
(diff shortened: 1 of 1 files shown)


phillimorland (approved):
super cool, as for location question, I don't know advantages or issues of a separate file, @fuzzylogic2000 @rmader do you guys have any issue with this? if not would merge.

fuzzylogic2000 (approved):
I think the location is fine! As long as it's only one mixin and only used in the userdashboard, it's easier to find like that!

PR #65, opened 2021-01-26
Pm 2021 01 platform branding

phillimorland:
I feel like i slightly wrote this out the long way, happy to update if there is a better way to do it.
Wanted to add the address setting to the emails but can't with a richtext field, do we mind?


adhocracy-plus/templates/500.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n static %}
 <html>
 <head>
-    <title>Error &mdash; {% trans 'adhocracy+' %}</title>
+    <title>Error &mdash; {{ settings.a4_candy_cms_settings.OrganisationSettings.platform_name }}</title>
```
fuzzylogic2000, line 4: I am not sure we should use it in the 500 page? It might be broken. I'll try to check.

fuzzylogic2000 (approved):
Nice! I will check the 500 before I merge, though.

fuzzylogic2000:
> I feel like i slightly wrote this out the long way, happy to update if there is a better way to do it.
> Wanted to add the address setting to the emails but can't with a richtext field, do we mind?

I think, this is the way to do it! I think, the address is fine: it is likely that we still do operate it and otherwise, people can find it and change it.

adhocracy-plus/templates/500.html
fuzzylogic2000: Alright, I didn't find out. Even on master, when I broke the base template, there was no styling. 
(I did:
set DEBUG=False and add ALLOWED_HOSTS = ['localhost'] in/to settings/dev.py
start server with: python manage.py runserver 8004 --insecure 
also started the npm and changed a template tag in the base.html)
So, I will merge now and think about it again another time.

PR #67, opened 2021-01-28
userdashboard: add skeleton of react component

phillimorland:


fuzzylogic2000 (commented):
Very nice! I will go on with this!

fuzzylogic2000:
This should have been a comment a while ago (I confusingly put it into the description): 
So, I totally broke the react. It's not rendering and not passing the linter and nothing. Will look at it again tomorrow.

PR #73, opened 2021-02-16
Kl 2021 02 helptexts

### LINKED ISSUE #69: Wagtail questions: Naming "organisation" and where to find adress
**URL:** https://kosmo-dev.liqd.net/admin/settings/a4_candy_cms_settings/organisationsettings/2/
**user:** admin

**Comment/Question:** Whoop whoop, thank you so much for the adaptation of the platform name! I have only questions:

- Where are "adress" and "contacts" of the organisation shown? 
- I also found the naming "organisation settings" a little confusing, as we also have the organisation dashboard with the projects - but this would be even one level higher. I think we should make the differences clear - maybe by claiming them "platform settings"? What do you think?
 
Bildschirmfoto 2021-02-01 um 12 30 11

fuzzylogic2000:
fixes #69


apps/cms/settings/models.py
```diff
@@ -91,15 +91,24 @@ class ImportantPages(BaseSetting):
...
+        help_text="The address is published on the contact form."
     )
-    address = fields.RichTextField()
-    contacts = fields.RichTextField()
+    contacts = fields.RichTextField(
+        help_text="The contacts are published on the contact form."
+    )
+
+    class Meta:
+        verbose_name = 'Platform settings'
```
Rineee, line 111: Should we make it translatable (also for important pages)? Because the rest of the settings options show in German, I suppose coming from wagtail?

fuzzylogic2000: Yeah, I also thought about that, but all the wagtail we added, also in the page models wasn't translated, so I didn't do that.

Rineee: Ok. Just thought it looks weird in the menu with a mix of German and English, but as its a mix everywhere, probably it doesnt matter.

Rineee:
Cool!

PR #74, opened 2021-02-24
Pm 2021 02 moderation tab 2

phillimorland:
react component displaying and all required data is in state, will need a refactor but this fulfills story for now.


phillimorland:
@fuzzylogic2000 I left in the commit where I changed how component was loaded so you can see difference

fuzzylogic2000 (approved):
Whoop whoop! Very nice! 
Do you want to squash the commits before merging? I think that would be good for cherry-picking to a+ at some point. :)

fuzzylogic2000:
I just put it all in one commit! :)

PR #75, opened 2021-03-23
templates/500.html: rm setting platform name and replace and add inli…

phillimorland:
…ne css to ensure it's styled when error occures

To test: 
- In requierements/base.txt change 
    ```whitenoise==3.3.1```
- In settings/dev.py change 
    ```set DEBUG=False``` 
and add
    ```ALLOWED_HOSTS = ['localhost']```
- In terminal start server with: 
    ```python manage.py runserver --insecure```
- In base.html remove a block tag to trigger error


phillimorland:
I will update the ribbon colours when I get them for the release story so if name is wrong, can update it then too.

fuzzylogic2000:
Ok, after messing around with git even more, I cherry-picked this from our old master (??) and it's in another PR now. Will just close this and hope for the best!

PR #77, opened 2021-03-24
Pm 2021 03 rebase aplus take 2

phillimorland:
I've tested maps and interactive event to check js elements and it all seems to be working.


Rineee:
I just checked moderation tab and there are no projects shown I think?

Rineee:
Yeah, some stuff is not loaded:
image

phillimorland:
@Rineee Mattermost is down so not sure if you saw, all my projects are listed in mod tab and I have cached js on aplus too so didn't think it was an issue but as said will look at it in the morning, if you commit what you did in a seperate commit and i'll take a look then :)

phillimorland:
@Rineee you were right about the mod dashboard entrypoint, mine must have been cached, have re-added. Network tab shows 302 on a+ too but on a force refresh it goes away on both so I think it's fine.

fuzzylogic2000:
Nice! But we don't merge this, but force push on master, right? And keep this branch or the old master or both as backup or how would we do that?

PR #78, opened 2021-03-25
WIP apps: add classifications app with inital models

Rineee:
@fuzzylogic2000 What do you think about this model structure? Or we just use the same for user and AI and have an AI_user or something..


fuzzylogic2000 (approved):
Yes, I do like it like that! Everything that we might need to change about that later (different choices for both models, generic foreign key, and the text), we can still change later! Cool! Let's do it like that and see how far it takes us. :)

PR #84, opened 2021-03-26
apps: add classifications app with inital AIClassification model

Rineee:


apps/classifications/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+"""
+We use the a4 reports for user classifications for now
+until we decided on a final reporting/classification procedure.
+Since we would have to disconnect reports and instead connect
+UserClassification to the comments, this seemed too complicated
+without a clear goal.
+The reports can then be shown in moderation dashboard without
+a classification.
```
fuzzylogic2000, line 40: Yes! Good to put it here and like that!

fuzzylogic2000 (approved):
Cool! I'll wait for emergency standup before merge, though!

phillimorland (approved):
Looks like a nice solution!

PR #85, opened 2021-04-01
Pm 2021 03 mod dashboard 2

phillimorland:
This is a refactor of the react component and introducing polling so it auto updates


fuzzylogic2000 (approved):
Supernice!

PR #86, opened 2021-04-01
project/serializer//userdashboard/ModerationProjects: adding addition…

phillimorland:
…al info for mod dashboard and adding it to component

just added dummy numbers for comment count and offensive so far

### DIFF
--- a/apps/projects/serializers.py
+++ b/apps/projects/serializers.py
@@ -1,3 +1,6 @@
+from functools import lru_cache
+
+from django.utils.translation import ugettext as _
 from easy_thumbnails.files import get_thumbnailer
 from rest_framework import serializers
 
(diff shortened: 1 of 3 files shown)


phillimorland:
ok thats a big drop, will look into tests next week!

phillimorland:
@Rineee actually got this working so feel ree to check it! :)

apps/projects/serializers.py
```diff
@@ -10,11 +13,47 @@ class ProjectSerializer(serializers.ModelSerializer):
...
-                  'tile_image_copyright']
+                  'tile_image_copyright',
+                  'status', 'access',
+                  'participation_active',
+                  'participation_string',
+                  'future_phase', 'active_phase',
+                  'past_phase',
+                  'offensive', 'comment_count']
+
+    @lru_cache(maxsize=1)
```
Rineee, line 36: What does this do? Cache the status somehow?

Rineee:
Cool, its all there and working and the tests even increased :+1: 
It looks a bit squeezed though for me, is that just for later and I merge it like this?
image

apps/projects/serializers.py
phillimorland: This is an opteermising cache algorithum used in mB which I just copied over with some of the serializer stuff, it removes least recently used cached items, I would imagine we won't have long project lists for the mods but I would say it doesn't hurt to have it, what do you think?

phillimorland (commented):
which broswers? mines looking bit different to that

Rineee:
> which broswers? mines looking bit different to that

In firefox. I can also merge first, whatever you prefer!

apps/projects/serializers.py
Rineee: Sure, we can keep it. Was just wondering what it does, but sounds good!

phillimorland:
@Rineee try merging? I can't reproduce currently

PR #90, opened 2021-04-07
Ks 2021 04 add helper to get all comments in project

Rineee:
There are some helper functions that are currently not used, but I left them in because I think we might need them in future..


phillimorland (approved):
super cool!

PR #91, opened 2021-04-07
Ks 2021 add classification when report is sent

Rineee:
Found a way to save a user classification every time a comment is reported without touching the reporting procedure (report is still sent). Next step would then be to adapt the report modal such that users can choose the classification (currently its always offensive).


phillimorland (approved):
Very cool!

PR #93, opened 2021-04-08
Fixing filter 'select affiliation'

### LINKED ISSUE #92: Interactive Events Module: Selecting affiliation option "all" is not working as expected.
**URL:** http://localhost:8004/liqd-orga/projects/module/interaktive-veranstaltung/ (to be replicated on kosmo-dev)
**user:** any
**expected behaviour:** When clicking on "Going to Event", And selecting "select affiliation", Then choosing "All", Should show "all" as selected
**behaviour:** However instead of "all", its showing "select affiliation"
**important screensize:** any
**device & browser:** any
**Comment/Question:** 

Screenshot?
select_dropdown

khamui:
Creating an array of affiliation categories and then add to
this array a default option 'all'. Then create a map of JSX from each item
including 'all' item.

Fixes #92

### DIFF
--- a/apps/interactiveevents/assets/js/a4_candy_interactive_events/livequestion_box.jsx
+++ b/apps/interactiveevents/assets/js/a4_candy_interactive_events/livequestion_box.jsx
@@ -79,7 +79,8 @@ export default class QuestionBox extends React.Component {
   }
 
   isInFilter (item) {
-    return (this.state.category === '-1' || this.state.category === item.category) &&
+    const { category } = this.state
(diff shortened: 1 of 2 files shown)


phillimorland (changes requested):
nice and clean, however when selecting 'all' then list should show all categories however as it is appended to the list then when 'all' is selected the list doesn't show any questions

khamui:
good catch! should be fixed with last commit.

phillimorland (approved):
super cool!

PR #94, opened 2021-04-08
apps/userdashboard: adding moderation detail page with dummy data - i…

phillimorland:


phillimorland:
I think as there is no link through to the detail page we could probably already merge this?

apps/userdashboard/assets/js/a4_candy_userdashboard/react_moderation_comment_list.jsx
```diff
@@ -0,0 +1,18 @@
+import ModerationCommentList from './ModerationCommentList'
+import React from 'react'
+import ReactDOM from 'react-dom'
+
+function init () {
+  $('[data-aplus-widget="moderation_comment_list"').each(function (i, element) {
```
khamui, line 6: using vanilla JS to select the <div data-aplus-widget> instead of jQuery.
Something like that:

.querySelectorAll('div[data-aplus-widget]')

khamui (commented):
please see my comments :)

fuzzylogic2000 (approved):
Cool! Looks nice!
Should we change the jQuery-thing to first merge this and then rebase #96 ?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationCommentList.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import django from 'django'
+
+import ModerationComment from './ModerationComment'
+
+export default function (props) {
+  const byText = django.gettext('By ')
+
+  return (
+    <div className="row mb-2">
```
khamui, line 10: We could handle the list here (loop/map).

apps/userdashboard/assets/js/a4_candy_userdashboard/react_moderation_comment_list.jsx
khamui: missing bracket :)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationCommentList.jsx
khamui: In case we keep the functional component here, we could give it a name and export it afterwards. Something like:

`function moderationCommentList (props) { ... }` or `const moderationCommentList = (props) => { ... }`

then

`export default moderationCommentList`

PS: i know it looks strange with the small caps name, but when importing we can do `ModerationCommentList`. At least we have a name for the function, no?

khamui (commented):
Sorry yesterday i mixed the commits over two commits. I just removed them on the other one and added them here.
Please, let me know if something is unclear.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationCommentList.jsx
phillimorland: @khamui so we use camel case for functional components and pascal for class components?

phillimorland: ok after swapping this I found our linting will not allow unused variables and when we don't need export it then adding a name goes against linting so what i'll do is add this to our design learnings and maybe we could consider a linting update if we feel it's too unreadable.

PR #95, opened 2021-04-09
Ks 2021 04 send comments to ai

Rineee:
I also opened a PR in admin repo with the auth token, so to test you can copy the token from there to settings.

### DIFF
--- a/apps/classifications/signals.py
+++ b/apps/classifications/signals.py
@@ -1,0 +1,51 @@
+import logging
+
+import backoff
+import httpx
+from django.conf import settings
+from django.db.models import signals
+from django.dispatch import receiver
+
+from adhocracy4.comments.models import Comment
+
+from .models import AIClassification
+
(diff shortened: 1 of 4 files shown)


apps/classifications/signals.py
```diff
@@ -0,0 +1,51 @@
...
+        response = call_ai_api(comment=instance)
+        if response.status_code == 200 \
+                and response.json()['classification'] == 'OFFENSE':
+
+            classification = AIClassification(
+                comment=instance,
+                classification='OFFENSIVE')
+            classification.save()
+    except httpx.HTTPError:
+        logger.error('Error connecting to %s', settings.AI_API_URL)
```
Rineee, line 29: I am hoping that this way the error will appear in sentry, but not sure if this works?

fuzzylogic2000 (approved):
Very nice! That you also thought about retriying when the API isn't available. :)

PR #96, opened 2021-04-12
Ks 2021 04 pipe through reported comments

Rineee:
A added the project slug to the url, so moderation detail is now available under userdashboard/moderation/detail/project_slug

It now only gets the comments that have been reported


fuzzylogic2000 (commented):
Nice!
Now we still need to

- [ ] get the comment info into react
- [ ] get the classified instead of the reported comments
- [ ] make the link in the moderation overview go to the detail

Did I forget anything?

fuzzylogic2000:
Haha, great, I only forgot the actual API. :p

Rineee:
Closed in favor of https://github.com/liqd/a4-kosmo/pull/111

PR #99, opened 2021-04-13
Ks 2021 04 add test for ai communication

Rineee:
To make this the signal test work locally, you need to insert the api token from admin repo into settings again.. maybe we can just remember that this fails locally? Or do you have any ideas how else to do it?

### DIFF
--- a/tests/classifications/test_classification_signals.py
+++ b/tests/classifications/test_classification_signals.py
@@ -1,0 +1,15 @@
+import pytest
+from django.db.models import signals
+
+from adhocracy4.comments.models import Comment
+from apps.classifications.signals import get_ai_classification
+
+
+@pytest.mark.django_db
+def test_ai_classification_receives_post_save_comment_signal(idea,
+                                                             comment_factory,
+                                                             caplog):
+    assert(get_ai_classification in signals.post_save._live_receivers(Comment))
+    comment_factory(content_object=idea,
+                    comment='lala')
+    assert('Error connecting to %s' in str(caplog.records[-1]))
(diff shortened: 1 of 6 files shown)


Rineee:
> To make this the signal test work locally, you need to insert the api token from admin repo into settings again.. maybe we can just remember that this fails locally? Or do you have any ideas how else to do it?

Ok no, it also fails here.. because it doesnt have the local.py settings I guess? Dont really know how to deal with it then.. do you have any idea @fuzzylogic2000 ?

fuzzylogic2000:
I think, we need to fake the API call instead of trying it. Otherwise we would be also testing the API, but only want to test if the answer we get triggers/doesn't trigger saving a AIclassification instance.

tests/classifications/test_classification_signals.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@pytest.mark.django_db
+def test_ai_classification_receives_post_save_comment_signal(idea,
+                                                             comment_factory,
+                                                             caplog):
+    assert(get_ai_classification in signals.post_save._live_receivers(Comment))
+    comment_factory(content_object=idea,
+                    comment='lala')
+    assert('Error connecting to %s' in str(caplog.records[-1]))
```
Rineee, line 15: So this is now just testing if the signal is received and that the ai call fails.. not optimal, but didnt find any other way to test it..

fuzzylogic2000: This might give a hint on how to do it: https://realpython.com/testing-third-party-apis-with-mocks/
But having the tests you wrote is already awesome, so I will merge!

PR #110, opened 2021-04-16
Kl 2021 04 translations

fuzzylogic2000:
This adds fork-specific translations and marks the strings that are fork-specific, so that we can only translate these in the locale-fork and don't overwrite translations we get from transifex.


fuzzylogic2000:
I tried to explain in the docs, waht I did: https://github.com/liqd/adhocracy-plus/pull/1217

phillimorland (approved):
super cool! I added a commit to rm obsolete make po-fork description and will cherry pick over

PR #111, opened 2021-04-16
Ks 2021 04 add classification serializer and api

Rineee:
There are several to-dos:

- [x] wasnt sure what data exactly we need to serialize, so this is a bit under construction -> thought to talk to @phillimorland about it on Monday?
- [ ] Hope I didnt mess up the react part :grimacing: There is also a lot to do still, I guess. Maybe we should also confirm again with MK what she wants or are we clear on that? Eg how the user message for user classifications should be shown. AIClassifications arent yet shown, but are there..
- [x] the link to the moderation detail is still missing, wasnt sure where to link to it from the project tile and where to link to the actual project page? Question for MK also?
- [x] the permission to view the api and/or moderation detail page for a certain project doesnt work for admins and I dont get why.. it should just check if admin is moderator of project, but it never gets there.. works for normal users though. Maybe Katharina can help?
- [ ] I realized that everything gets pretty slow when you are on moderation tab.. maybe we should reduce polling, what do you think? Or will this not be a problem on the server?

Not sure if we should merge and go on from there? Up to youu :)

### DIFF
--- a/apps/classifications/api.py
+++ b/apps/classifications/api.py
@@ -1,0 +1,53 @@
+from django.shortcuts import get_object_or_404
+from rest_framework import mixins
+from rest_framework import viewsets
+
+from adhocracy4.api.permissions import ViewSetRulesPermission
+from adhocracy4.projects.models import Project
+from apps.classifications.models import AIClassification
+from apps.classifications.models import UserClassification
+from apps.projects import helpers
+
+from . import serializers
+
+
+class ClassificationViewSet(mixins.ListModelMixin,
+                            viewsets.GenericViewSet):
+
+    permission_classes = (ViewSetRulesPermission,)
+
+    def dispatch(self, request, *args, **kwargs):
+        self.project_pk = kwargs.get('project_pk', '')
+        return super().dispatch(request, *args, **kwargs)
+
+    @property
+    def project(self):
+        return get_object_or_404(
+            Project,
+            pk=self.project_pk
+        )
+
+    def get_permission_object(self):
+        return self.project
+
+
+class UserClassificationViewSet(ClassificationViewSet):
+
+    serializer_class = serializers.UserClassificationSerializer
+
+    def get_queryset(self):
+        all_comments_project = helpers.get_all_comments_project(self.project)
+        return UserClassification.objects.\
+            filter(comment__in=all_comments_project)\
+            .order_by('created')
+
+
+class AIClassificationViewSet(ClassificationViewSet):
+
+    serializer_class = serializers.AIClassificationSerializer
+
+    def get_queryset(self):
+        all_comments_project = helpers.get_all_comments_project(self.project)
+        return AIClassification.objects.\
+            filter(comment__in=all_comments_project)\
+            .order_by('created')
--- a/apps/classifications/serializers.py
+++ b/apps/classifications/serializers.py
@@ -1,0 +1,82 @@
+from django.utils.translation import ugettext_lazy as _
+from easy_thumbnails.files import get_thumbnailer
+from rest_framework import serializers
+
+from adhocracy4.comments.models import Comment
+from apps.classifications.models import AIClassification
+from apps.classifications.models import UserClassification
+
+
+class CommentSerializer(serializers.ModelSerializer):
+
+    comment_url = serializers.SerializerMethodField()
+    user_name = serializers.SerializerMethodField()
+    user_image = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Comment
+        fields = ['comment_url', 'user_name', 'user_image']
+
+    def get_comment_url(self, instance):
+        return instance.get_absolute_url()
+
+    def get_user_name(self, instance):
+        if instance.is_censored or instance.is_removed:
+            return _('unknown user')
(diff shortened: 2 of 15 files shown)


fuzzylogic2000:
Will this replace #96 or is it extra?

Rineee:
> Will this replace #96 or is it extra?

Yes, it replaces it. I took some parts of it, but now the comments arent passed through context, but fetched via api, this is the way we wanted it, right?

apps/userdashboard/routers.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import routers
+
+from adhocracy4.api.routers import CustomRouterMixin
+
+
+class ModerationDetailRouterMixin(CustomRouterMixin):
+
+    prefix_regex = (
+        r'userdashboard/moderation/(?P<project_pk>[\d]+)/{prefix}'
```
fuzzylogic2000, line 9: What is the prefix for?

fuzzylogic2000 (commented):
Yes, nice! That's how we want it!

apps/userdashboard/routers.py
Rineee: Its to include the project_pk in the routes, so you can for example fetch /userdashboard/moderation/1/userclassifications list. Then project_pk=1 is in kwargs of the viewset. Its like its done for example in OrganisationDefaultRouter or ModuleDefaultRouter.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationCommentList.jsx
```diff
@@ -1,18 +1,69 @@
...
+    const userClassifications = this.state.userClassifications
+    const byText = django.gettext('By ')
+
+    return (
+      <div className="row mb-2">
+        <div className="col-12">
+          <h1 className="m-0">project title</h1>
+          <span className="text-muted">{byText}organisation</span>
+          {userClassifications.map((item, i) => (
+            <li key={i}>
```
phillimorland, line 59: @Rineee the <li> element is what is causing the bullet point for each comment as it assumes it's a list element so add default styling, we probably should do some juggling between this and the comment component maybe just make comment a div and rm <ul>and<li> and wrap the <li> elements in this component in a <ul>

apps/userdashboard/routers.py
fuzzylogic2000: And the prefix itself after the project pk? That's the UserClassification or AIClassification, then, right?

apps/classifications/api.py
```diff
@@ -0,0 +1,65 @@
...
+from adhocracy4.api.permissions import ViewSetRulesPermission
+from adhocracy4.projects.models import Project
+from apps.classifications.models import AIClassification
+from apps.classifications.models import UserClassification
+from apps.projects import helpers
+
+from . import serializers
+
+
+class UserClassificationViewSet(mixins.ListModelMixin,
```
fuzzylogic2000, line 14: As only the upper part is different for both ClassificationViewSets, you could make one, both inherit from. Or maybe that's only, when you use the code the third time. :)

apps/classifications/serializers.py
```diff
@@ -0,0 +1,81 @@
...
+        return instance.created.strftime('%d.%m.%y')
+
+
+class AIClassificationSerializer(serializers.ModelSerializer):
+    comment = CommentSerializer(read_only=True)
+    classification = serializers.SerializerMethodField()
+    created = serializers.SerializerMethodField()
+
+    class Meta:
+        model = UserClassification
```
fuzzylogic2000, line 74: AIClassification!

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationComment.jsx
```diff
@@ -2,58 +2,58 @@ import React, { Component } from 'react'
...
-      project: this.props.key
-    }
-  }
-
   render () {
+    const { classification, commentText, commentUrl, created, userImage, userName, aiClassified } = this.props
     const postedText = django.gettext(' posted a ')
-    const offensiveText = django.gettext(' for being offensive')
-    const categoryText = django.gettext('Categories: ')
+    const offensiveTextReport = django.gettext(' that has been reported as ')
```
fuzzylogic2000, line 8: Uh, I guess, because of the translations rebasing might stop here. You just need to use 
`django.pgettext('kosmo', ' that has been...` for all the strings, then.

fuzzylogic2000 (approved):
Nice! I think, we can merge and improve from there!

apps/classifications/serializers.py
Rineee: Good catch, thanks!

apps/classifications/api.py
Rineee: Sure, can do that

apps/userdashboard/routers.py
Rineee: Oh sorry, that didnt answer your question :see_no_evil: But yes exactly, its the prefix of the urls registered for this router, so userclassifications or aiclassifications

fuzzylogic2000 (approved):
Looking good!

Issue #72, opened 2021-02-11
UX/UI testing story https://taiga.liqd.net/project/liqd-liquid-software/us/3711


mcastro-lqd:
**URL:** https://kosmo-dev.liqd.net/userdashboard/overview/
**user:** *moderator/ initiator*
**important screensize:** *Both*
**device & browser:** *Safari Version 14.0.2 (16610.3.7.1.9)*

Desktop: 
[200211-d-3711-overview-of-users-dashboard.pdf](https://github.com/liqd/a4-kosmo/files/5964081/200211-d-3711-overview-of-users-dashboard.pdf)

Mobile:
[200211-m-3711-overview-of-users-dashboard.pdf](https://github.com/liqd/a4-kosmo/files/5964084/200211-m-3711-overview-of-users-dashboard.pdf)

phillimorland:
@mcastro-lqd thanks for these! I have some clarifying texts below also when looking through the designs I got a little confused about which ones, is there anyway to makr the final design on zeplin? maybe they can all have like a final tag on them? 
Desktop
5. I wasn't sure which design you were referring to about the current tiles? 

Mobile:
4. I cannot replicate this issue. Can I ask just to confirm, there is a list of organisations when viewing on desktop and when that list turns into a dropdown on mobile you cannot see any organisations even after opening and closing the dropdown?
6. style mobile arrows with zeplin design (check left cropping)
8. change tiles on view more screen? which desktop design? this referes to desktop number 5 i think, i could only find a comment here but nothing else although some of the design looked really old so I don't really want to be trying to copy anything thats not right https://app.zeplin.io/project/5fe363c415be15a48946ca29/screen/5fe369b5adeaaeacadb14449

mcastro-lqd:
@phillimorland Sorry, I missed these comments

Here you go:

5. Like the Profile one: 
<img width="1401" alt="Screenshot 2021-04-13 at 12 17 06" src="https://user-images.githubusercontent.com/59610786/114537162-2bbd2700-9c52-11eb-9f65-c3f6c27ad41b.png">

4. Yes, I just tested it again but I still can't see them. (Tested on Safari)
![IMG_9568](https://user-images.githubusercontent.com/59610786/114537691-cd447880-9c52-11eb-917f-d27c641d6438.jpg)

6. I just thought about making these cropped tiles that already gives the user the impression of a slidable gallery. Like these examples: 
[...]

PR #113, opened 2021-04-16
Kl 2021 04 follow overview

fuzzylogic2000:
The first commit fixes what I didn't notice when I merged #106 
I don't know if that tile was also meant to be replaced, though. @phillimorland do you? Was it meant for both the user dashboard overview and the user profile? And should both only display 6 tiles?


phillimorland:
@fuzzylogic2000 in the issue, it was requested that we use original tiles and as they don't work at width of 1/3 I changed amount https://github.com/liqd/a4-kosmo/issues/72#issuecomment-818632052

phillimorland (approved):
Nice!

PR #117, opened 2021-04-23
Ks 2021 04 add tests for classification api

Rineee:


fuzzylogic2000 (approved):
Very nice and good catch! Writing tests: <3

PR #121, opened 2021-04-26
Ks 2021 04 add use ai setting

Rineee:
@fuzzylogic2000 I now set the default to not use the ai, thus I would then add it to local settings for dev on pillar to use the ai. Or do you prefer the other way around?

There is a PR for adding the setting on pillar: https://github.com/liqd/admin/pull/287

### DIFF
--- a/adhocracy-plus/config/settings/base.py
+++ b/adhocracy-plus/config/settings/base.py
@@ -544,3 +544,4 @@ SITE_ID = 1 # overwrite this in local.py if needed
 
 AI_API_AUTH_TOKEN = ''
 AI_API_URL = 'https://kosmo-api-dev.liqd.net/api/classify/'
+AI_USAGE = True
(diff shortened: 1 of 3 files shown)


adhocracy-plus/config/settings/base.py
```diff
@@ -544,3 +544,4 @@
 
 AI_API_AUTH_TOKEN = ''
 AI_API_URL = 'https://kosmo-api-dev.liqd.net/api/classify/'
+AI_USAGE = False
```
fuzzylogic2000, line 547: I think, instead of adding this to pillar, we could set it to True here. Then if we deploy and forget to add the token to the pillar, we will get a warning!

fuzzylogic2000 (commented):
Supernice! But should we enable the API here and only add the token to the pillar?

adhocracy-plus/config/settings/base.py
Rineee: Right!

fuzzylogic2000 (approved):
Whoop whoop!

PR #125, opened 2021-04-27
Ks 2021 04 user interplated translations

Rineee:
I changed the classification choices to lowercase in the model.. sorry about that, should have done when adding it. But now it was annoying, because I needed it to be displayed lower case (which I did before with styling, but removed it now, because in german we need upper case also) and just changing it at some other point would make the translations not work anymore, because the translated string was upper case... probably this is even more confusing, sorry /:


fuzzylogic2000 (approved):
Looks good!
Should we also add the new translation strings in this PR?

PR #131, opened 2021-08-06
docs/installation_prod: adapt to kosmo

Rineee:
There are things that Im not sure of, see comments... but once that is clear, I would give this file to @hmkay as a installation manual?

### DIFF
--- a/docs/installation_prod.md
+++ b/docs/installation_prod.md
@@ -19,21 +19,21 @@ This guide will focus on Debian/Ubuntu-like systems - however, it works very sim
 Create and switch to user (as `root` or using `sudo`)
 
 ```
-adduser aplus
-su aplus
+adduser kosmo
+su kosmo
 cd
 ```
 
 #### Get the code
 ```
-git clone https://github.com/liqd/adhocracy-plus.git
-cd adhocracy-plus
+git clone https://github.com/liqd/a4-kosmo.git
+cd a4-kosmo
 git checkout release
 ```
 
 #### Create and launch virtual environment
 ```
-mkvirtualenv --python=/usr/bin/python3 aplus
+mkvirtualenv --python=/usr/bin/python3 kosmo
 ```
 
 Note: you won't need the `--python` part when using a recent distribution.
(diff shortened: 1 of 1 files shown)


docs/installation_prod.md
```diff
@@ -119,65 +119,65 @@ Cancel the server after testing via `ctrl`+`c`
...
+Description=kosmo server
 After=network.target
 
 [Service]
-User=aplus
-WorkingDirectory=/home/aplus/adhocracy-plus
-ExecStart=/home/aplus/.virtualenvs/aplus/bin/gunicorn -e DJANGO_SETTINGS_MODULE=adhocracy-plus.config.settings.production --workers 4 --threads 2 -b 127.0.0.1:8000 -n adhocracy-plus adhocracy-plus.config.wsgi
+User=kosmo
+WorkingDirectory=/home/kosmo/a4-kosmo
+ExecStart=/home/kosmo/.virtualenvs/kosmo/bin/gunicorn -e DJANGO_SETTINGS_MODULE=adhocracy-plus.config.settings.production --workers 4 --threads 2 -b 127.0.0.1:8000 -n kosmo adhocracy-plus.config.wsgi
```
Rineee, line 132: here Im not sure what the -n option is? maybe @goapunk can tell me?

docs/installation_prod.md
```diff
@@ -195,18 +195,18 @@ server {
 
   server_name your.domain;
 
-  # forward traffic to adhocracy-plus
+  # forward traffic to kosmo
   location / {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
     proxy_set_header Host $http_host;
     proxy_pass http://127.0.0.1:8000;
```
Rineee, line 203: Do I need to change something here? @goapunk

docs/installation_prod.md
goapunk: -n is the process name gunicorn will use (it requires the python setproctitle package):
e.g. without -n it will look like this in the list of running processes:
```
ps ax | grep kosmo
575 ?        Sl     0:09 bin/python3 /home/django/a4_kosmo/bin/gunicorn -e DJANGO_SETTINGS_MODULE=adhocracy-plus.config.settings.production_default --workers 4 --threads 2 -b 127.0.0.1:6000 -n a4_kosmo adhocracy-plus.config.wsgi
```
with -n it looks like this:
```
ps ax | grep kosmo
  12855 ?        S      0:00 gunicorn: master [a4_kosmo]
```

docs/installation_prod.md
goapunk: no, this should be fine, it just needs to be the same ip:port as used above in the `ExecStart` line for gunicorn.

docs/installation_prod.md
Rineee: Ah cool, thanks for the explanation! So naming it kosmo makes sense I guess?

goapunk: yup, that's fine I'd say

fuzzylogic2000 (approved):
Very nice!

PR #139, opened 2021-11-02
apps/classification: add is_pending

fuzzylogic2000:


apps/classifications/rules.py
```diff
@@ -1,9 +1,16 @@
...
 from adhocracy4.projects.predicates import is_moderator as is_project_moderator
 
 rules.add_perm('a4_candy_classifications.view_userclassification',
                is_project_moderator)
 
 rules.add_perm('a4_candy_classifications.view_aiclassification',
                is_project_moderator)
+
+rules.add_perm('a4_candy_classifications.change_userclassification',
+               module_predicates.is_context_moderator)
```
fuzzylogic2000, line 13: using this rule produced
```
{"detail":"You do not have permission to perform this action."}
```

while using the other one (`is_project_moderator`) gave a real long error (try it!) :)

fuzzylogic2000 (commented):
I tried this locally with 
```
curl -d '{"is_pending":False}' -H "Content-Type: application/json" -X PATCH http://localhost:8004/api/userdashboard/moderation/4/aiclassifications/1/ --user 'admin@liqd.net':'password'
```
(replace module and classification id)

apps/classifications/rules.py
Rineee: So I actually think you got it all right, but the token authentication thing from Aplus got in the way.. I didnt find out how to circumvent that with curl, but if you call /api/login first with the normal credentials, you get a token and can use that with curl...

We need to use the module_predicates.is_context_moderator here, because for the patch request, the permission_object is the classification (that happens in ViewSetRulesPermission again :)) and not the project..

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationComment.jsx
```diff
@@ -13,6 +22,29 @@ export default class ModerationComment extends Component {
...
+      .then(res => res.json())
+      .then(json => {
+        this.setState({
+          isPending: json.is_pending
+        })
+      }).catch((err) => {
+        console.log(err)
+      })
+  }
+
```
Rineee, line 47: We should probably think about having a central API file somewhere like in Z:T?

Rineee, line 47: We should probably think about having a central API file somewhere like in Z:T?

Rineee:
@fuzzylogic2000 I couldnt add you as a reviewer, because its your PR, but please do so :)

phillimorland:
@Rineee I changed the dropdown a little to align with designs but this is what is in zeplin now except the icon which we'll need from Marcelo

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -13,6 +23,36 @@ export default class ModerationNotification extends Component {
...
+        this.setState({
+          isPending: json.is_pending
+        })
+      }).catch((err) => {
+        console.log(err)
+      })
+  }
+
+  toggleIsBlocked () {
+    const newIsBlocked = !this.state.isBlocked
```
sabinammm, line 50: Just for my understanding:
this is in reference to the comment variable **is_blocked** comiing from a4, right? And will this be added to the ModerationNotificationList.jsx and changed accordingly, which is a similar way that blocked comments will then not be shown in discussion? (I'm going to try this next and still aiming to grasp the setup)

fuzzylogic2000: The first part: yes. @phillimorland just added a placeholder which will be filled once the actual field is merged.
And the second part I do not quite get. Here it's a button but for the discussion, we would either not show them or show them differently depending on the value of is_blocked.

fuzzylogic2000 (commented):
So cool! 
But after archiving the notification tile says, that the comment was blocked. We can set that straight when we also have the is_blocked though and just merge this like it is?

Rineee:
> So cool! But after archiving the notification tile says, that the comment was blocked. We can set that straight when we also have the is_blocked though and just merge this like it is?

@phillimorland I tried to solve this in the last commit. But then the unarchive button was in the wrong place because of justify-content-between I think, so changed that to justify-content-end in that case. It is working, but I dont know if this is the right way to do it? To be safe, I added it in its own commit, so we can remove it again :)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -13,6 +22,29 @@ export default class ModerationNotification extends Component {
     )
   }
 
+  async switchIsPending () {
```
khamui: just a little note here, because i saw it in other spots already.

we should only use `async` if we use `await` in the function body. otherwise we are turning this function into a asynchronous one, but not really use it. so either `Promise: then` or `async/await`.

we could include a rule in eslint:
https://eslint.org/docs/rules/require-await

phillimorland (approved):
Perfect use of classes! :D

PR #140, opened 2021-11-02
Ks 2021 11 add archived notification filter

Rineee:
Ok, this was easier than I thought :)

@kha The api calls to get the classifications in ModerationCommentList need to be extended with the query parameter to be able to filter, so e.g.
`this.props.userclassificationApiUrl+'?is_pending=true'` to get only pending user classifications or `this.props.aiclassificationApiUrl+'?is_pending='` to get all (I guess this is the same as just leaving the query param out).

I also added the model stuff here so you could already work with it..


khamui:
@Rineee nice!! i like that with filter params.

Just to confirm it is the expected behavior.

if I send a request with `aiclassifications/?is_pending=true` or `aiclassifications/?is_pending` or even `aiclassifications/` i get the list of pending notifications.

if I send with `aiclassifications/?is_pending=0` or `aiclassifications/?is_pending=false` i do not receive them.

Is it intended to be like that?

Rineee:
@khamui Almost I think :) So there should be the possibility to get all classifications (meaning pending or not), this should be possible with `aiclassifications/?is_pending=` and probably also with `aiclassifications/` and  `aiclassifications/?is_pending`.
And then you should be able to only get pending ones: `aiclassifications/?is_pending=true` or not pending ones `aiclassifications/?is_pending=false`. To test that, after you ran the migration for is_pending, you can set is_pending=false in django admin for a classification.

khamui:
now i got it :) works like a charm

PR #141, opened 2021-11-03
Pm 2021 11 tile buttons

phillimorland:


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationComment.jsx
```diff
@@ -42,12 +44,31 @@ export default class ModerationComment extends Component {
...
+            <div className="text-end">
+              <button
+                type="button" className="dropdown-toggle btn btn--link" aria-haspopup="true"
+                aria-expanded="false" data-bs-toggle="dropdown"
+              >
+                <i className="fas fa-ellipsis-v" aria-hidden="true" />
+              </button>
+              <ul className="dropdown-menu dropdown-menu-end">
+                <li key="1">
+                  <button className="dropdown-item" type="button">{archiveText}</button>
```
phillimorland, line 63: @Rineee here is where you can add the archive action then I can take over again if you like and set correct buttons to show and when to enable/disable them, currently everything is shown

Rineee:
Looking great!

PR #143, opened 2021-11-04
#4925 adding just text component - no save or backend

phillimorland:
adds a basic textinput field below nottifications in the moderation dashboard, starts story #4925


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNote.jsx
```diff
@@ -0,0 +1,57 @@
...
+        objectPk: this.props.subjectId,
+        contentTypeId: this.props.subjectType
+      }
+    }
+    if (!note) {
+      return
+    }
+    this.props.onNoteSubmit(data, this.props.parentIndex).then(() => {
+      this.clearForm()
+    })
```
phillimorland, line 33: not needed

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNote.jsx
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from 'react'
+import django from 'django'
+
+export default class ModerationNote extends Component {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      note: '',
+      noteCharCount: 0
```
khamui, line 10: is this needed to be stored in state?
would `this.state.note.length` work wherever you need the charcount?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -126,7 +137,7 @@ export default class ModerationNotification extends Component {
         </div>
         <div className={'mt-3 d-flex justify-content-' + (!this.state.isPending && !this.state.isBlocked ? 'end' : 'between')}>
           {this.state.isPending
-            ? <><button className="btn btn--none" type="button"><i className="fas fa-reply" aria-hidden="true" />{replyText}</button>
+            ? <><button className="btn btn--none" type="button" onClick={this.toggleModerationNote.bind(this)}><i className="fas fa-reply" aria-hidden="true" />{replyText}</button>
```
khamui, line 140: any reason not to note this handler same as `this.toggleIsBlocked()` and `this.toggleIsPending()`?

something like `(e) => this.toggleModerationNote(e)` should work the same way.
I kind of avoid the `.bind(this)`, because it always looked to me like a workaround (i know it is legacy javascript style, but i think the arrow notation is more readable and we use it below).

khamui (commented):
works fine and looks pretty neat!!

just two comments, but nothing essential

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNote.jsx
phillimorland: I don't know I just copied the comment but I feel like it was something to do with .length not being a string?! but will try that when we come back to it, thank you!

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
phillimorland: first question: not that I know of.
second question: yeah I am having trouble adapting to arrow notation as I find it less readable, will work on it when we come back to it though, cheers!

khamui:
@phillimorland @fuzzylogic2000 any objections to merge this into main?

fuzzylogic2000:
> @phillimorland @fuzzylogic2000 any objections to merge this into main?

I thought "on hold" meant that we still wait with that until we come back to it? I don't know what the status is here, though. If you reviewed and it's fine, we can also merge and then go on when we come back to it.

phillimorland:
@khamui @fuzzylogic2000 as there is a release scheduled soon for KOSMO, I thought merging something that isn't working would be confusing for testing and we'll just get a bunch of unnecessry issues, hence `on hold`

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
phillimorland: @khamui see new PR https://github.com/liqd/a4-kosmo/pull/173

PR #144, opened 2021-11-08
tests/projexct api: rename tests - fixes #136

### LINKED ISSUE #136: normal user can access moderation dashboard
**URL:** see test https://github.com/liqd/a4-kosmo/blob/main/tests/projects/test_project_api.py
**user:** 
**expected behaviour:** I am not sure if we should forbid this.
**behaviour:** Currently the user can access the moderation dashboard, but the right rule is used in the view. 
**important screensize:**
**device & browser:** 
**Comment/Question:** 
https://github.com/liqd/a4-kosmo/commit/dfde4ad27682142838f1ac0eeac01c73910d3eec 
https://github.com/liqd/a4-kosmo/blob/main/apps/userdashboard/rules.py

fuzzylogic2000:


Rineee (approved):
Yes, perfect!

PR #148, opened 2021-11-09
moderation/notifications/filter: fixing wrongly shown notifications w…

### LINKED ISSUE #147: Moderation Dashboard: Filtering (All, Pending, Archived) not immediate!
**URL:** local/dev
**user:** moderator
**expected behaviour:** Changing filter from "All" to "Pending" or "Archived" showing immediately the right content
**behaviour:**  Changing leads to -sometimes- showing wrong notifications. Being corrected after 1-2 seconds.
**important screensize:** any
**device & browser:** tested in chrome
**Comment/Question:** It filters correctly, but there is a glitch of ~1 sec. where it can show wrong notifications.

Screenshot?

khamui:
…hen filtering.

added a frontend filtering and making sure loading state changes after fetching

closes #147

### DIFF
--- a/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
+++ b/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
@@ -11,23 +11,23 @@ export default class ModerationNotificationList extends Component {
 
     this.state = {
       notifications: [],
-      filter: '',
+      filterItem: { filter: '', name: '' },
       isLoaded: false,
       alert: undefined
     }
   }
 
   componentDidMount () {
-    this.loadData(this.state.filter)
-    this.timer = setInterval(() => this.loadData(this.state.filter), 3000)
+    this.loadData(this.state.filterItem.filter)
+    this.timer = setInterval(() => this.loadData(this.state.filterItem.filter), 3000)
   }
 
-  filterChangeHandle (filter) {
+  filterChangeHandle (filterItem) {
     this.setState({
-      filter: filter,
+      filterItem: filterItem,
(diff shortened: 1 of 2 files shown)


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
```diff
@@ -44,7 +44,10 @@ export default class ModerationNotificationList extends Component {
     const userFetch = await fetch(userUrl)
     const userNotifications = await userFetch.json()
 
-    this.setState({ notifications: [...aiNotifications, ...userNotifications], isLoaded: true })
+    this.setState(
+      { notifications: [...aiNotifications, ...userNotifications] },
+      () => this.setState({ isLoaded: true })
```
khamui, line 49: this might look weird at first sight, i just read in the docs that you can add a callback as a second argument, which waits for the first argument (setting the notifications) to be finished. does that make sense?

sabinammm: It's still happening (but only every 20-30th time or so I switch). Also, it doesn't change after a few seconds.
image

 If I reload, it switches the filter to "All" and shows all notifications.

phillimorland: @sabinammm reload behaviour is expected else we would have to set cookies

phillimorland (approved):
@khamui just realised I can't reproduce this issue either on dev or locally on main (which is interesting) but code looks good so will merge and we can see how we get on

PR #149, opened 2021-11-09
Ks 11 2021 make comment blockable via api

Rineee:
only works with recent version of A4 with is_blocked in introduced


khamui:
works fine!

PR #150, opened 2021-11-09
Kt 2021 11 blocking comment moderation

khamui:


sabinammm (approved):
Awesome!! Works for me:)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -23,35 +23,44 @@ export default class ModerationNotification extends Component {
...
-  toggleIsBlocked () {
-    const newIsBlocked = !this.state.isBlocked
-    this.setState({
-      isBlocked: newIsBlocked
-    })
+  async toggleIsPending () {
+    const [response, error] =
+      await this.safeFetch({ is_pending: !this.state.isPending })
+    const alertMessage = response && response.is_pending
+      ? 'unarchived'
```
Rineee, line 46: I think this needs to be translatable as well?

Rineee:
Supercool! :tada: Just one little thing about translations...

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
khamui: that is true. i will think about a different solution, because using these translated words to identify is a bit too error prone.

khamui: @Rineee okay i shifted the translation of the word to the end as well, so i can keep these words as identifiers.

phillimorland (approved):
fabulous! :)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -23,35 +23,44 @@ export default class ModerationNotification extends Component {
...
+      : 'archived'
+
+    !error || this.props.onChangeBlocked(error)
+    !error && this.props.onChangeBlocked(alertMessage)
+    !error && this.setState({ isPending: response.is_pending })
+  }
```
phillimorland, line 52: @khamui can you explain this syntax? i'm not finding it very readable

khamui: yes i have to admit this is a very shorthand style of writing if statements. I find, if you get use to it, than this allows code to be placed close to each other to make the block even easy to read. Translated into normal if/else notation it would be (I can change that, but i think the extra lines of an if/else is disturbing somehow):

```
if (!error) {
  this.props.onChangeBlocked(alertMessage)
  this.setState({ isPending: response.is_pending })
} else {
  this.props.onChangeBlocked(error)
}
```

this i would need to write two times for `toggleIsPending` and `toggleIsBlocked`. I don't mind to change it...

khamui: oh the first comment was just the justification haha. the explanation of those lines is:

first part the statement above is the condition `!error` and the second part is the execution `this.props.onChangeBlocked(error)`.

In case of a `&&` operator it will execute if the `!error` is `true`. In case of `||` the second part will execute if `!error` is `false`.

phillimorland: I would take readability over less code personally but @Rineee @fuzzylogic2000 what are your thoughts? i'm happy to go with the majority?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
```diff
@@ -62,17 +63,18 @@ export default class ModerationNotificationList extends Component {
     })
   }
 
-  getArchivedAlert = () => {
+  getSuccessAlert = (message) => {
+    const alertMessage = `Notification ${django.gettext(message)} successfully.`
```
phillimorland, line 67: Sorry just saw this, so we run a translation on the `message` word and then we run a translation on the notification sentance with the `message` word in? is there a reason we don't just translate it once?

khamui: @phillimorland  reason for this nested translation is that i then do not need to have to distinguish between them. which would be a big if/elseif/else block with four times almost the same sentence.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
khamui: yes, you are right! i am going to change it, because most probably the both of us are the ones that have to change it in the future. lets go for the more convenient way.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
phillimorland: ahh cool, but we need to remember to check it when we run translation, we have had issues with strings in js files not being picked up in the past.

phillimorland (approved):
perfect! danke! :)

PR #173, opened 2022-04-04
#5761 react text component for moderation statement

phillimorland:
form component, no save as yet.

### DIFF
--- a/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
+++ b/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
@@ -1,6 +1,8 @@
 import React, { Component } from 'react'
-import Cookies from 'js-cookie'
 import django from 'django'
+import api from './api'
+
+import { ModerationStatementForm } from './ModerationStatementForm'
 
 export default class ModerationNotification extends Component {
   constructor (props) {
@@ -8,7 +10,8 @@ export default class ModerationNotification extends Component {
 
     this.state = {
       isPending: this.props.isPending,
-      isBlocked: this.props.isBlocked
+      isBlocked: this.props.isBlocked,
+      moderatorStatement: ''
     }
   }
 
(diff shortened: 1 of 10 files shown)


phillimorland:
@khamui I tried to update this to the current standard practice syntax, unlike the previous one, not sure if I got it all right though?!

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import Cookies from 'js-cookie'
 import django from 'django'
 
-import ModerationNote from './ModerationNote'
+import { ModerationStatement } from './ModerationStatement'
```
khamui, line 5: i guess this is the wording we use for this purpose right? `statement` not `feedback`?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationStatement.jsx
```diff
@@ -0,0 +1,40 @@
...
+import django from 'django'
+
+export const ModerationStatement = (props) => {
+  const [statement, setStatement] = useState(props.initialStatement)
+  const placeholderText = django.pgettext('kosmo', 'Write statement')
+  const charCountText = django.pgettext('kosmo', ' characters')
+
+  const handleSubmit = (e) => {
+    e.preventDefault()
+    const statement = this.state.statement.trim()
```
khamui, line 11: I believe `this` won't work here anymore.
`this.state.statement` is now simply `statement` i guess.

khamui: same for `this.props`, is now `props` in functional components

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationStatement.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react'
+import django from 'django'
+
+export const ModerationStatement = (props) => {
+  const [statement, setStatement] = useState(props.initialStatement)
```
khamui, line 5: in case `props.initialStatement` is `undefined` you might need to initialize `statement` with an empty string. otherwise the JSX below `statement.length` might break.

maybe something like `useState(props.initialStatement | '')`

khamui (commented):
Yay! just added some comments

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
phillimorland: I got statement from @fuzzylogic2000 comment [here](https://taiga.liqd.net/project/liqd-liquid-software/us/5761?milestone=302) I thought statement was more descriptive then feedback, what do you guys think?
Actually it could actually be `ModerationStatementForm` or `ModerationFeedbackForm`

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationStatement.jsx
phillimorland: ahh nice, I initially had it defined within the props parentheses but didn't like that and wasn't sure where else to put it, thank you!

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
fuzzylogic2000: Yes, as the model and the field are called statement, I think statement is good! 
```
class ModeratorCommentStatement(UserGeneratedContentModel):
    statement = RichTextField(
        blank=True,
        # Translators: kosmo
        verbose_name=_('Moderator feedback'),
    )
    comment = ...
```
(But the app is called moderatorfeedback as that already existed.)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationStatement.jsx
phillimorland: statement caused errors, will wait for api before i mess around here as I just straight up copied it from old component

apps/userdashboard/assets/js/a4_candy_userdashboard/api.js
```diff
@@ -0,0 +1,30 @@
+import Cookies from 'js-cookie'
```
khamui, line 1: @Rineee @fuzzylogic2000 @phillimorland I spent quite a bit of time thinking about that. but I think it is beneficial to extract that from the components. it makes them pretty messy if we leave it in there, especially if we have to add more different requests. so the best way i found, was to - again - put it in its own file, what do you think about and is there a more elegant way? I mean better would be to implement it in a4 but it's most probably not the right time?

Rineee (approved):
Cool, adding and deleting works, will add extra task for editing and merge now!

PR #174, opened 2022-04-05
apps/notifications/email: base email adjustments

khamui:
@phillimorland its still wip. I was wondering how to change the logo? I think the kosmo logo is a fallback logo?


phillimorland:
@khamui what do you mean by fallback? also did you try changing the file in the image folder? i think you'll need to do a make clean after and reinstall but it might work?! It's been a long while since I did email stuff

khamui:
> @khamui what do you mean by fallback? also did you try changing the file in the image folder? i think you'll need to do a make clean after and reinstall but it might work?! It's been a long while since I did email stuff

probably nothing relevant with the fallback. it seems if there is no other logo "provided/uploaded" it will use the Kosmo. but maybe feature from aplus.

fuzzylogic2000 (commented):
Is all of it specific to kosmo? Or are there bits that should be changed in a+ as well? In the second case doing the changes in a+ and rebasing will make things easier on the long run.

khamui:
@fuzzylogic2000 okay I just talked to @mcastro-lqd and for now this email template adjustments only affect Kosmo. Aplus mails remain as they are.

phillimorland (approved):
very nice!

PR #175, opened 2022-04-06
Ks 2022 04 use new classification categories

Rineee:
~~Do not merge until kosmo-api has been deployed!~~
Has been deployed, so this could be merged now.

To test you need to have newest main of kosmo-api (its here now: https://git.liqd.net/liqd/kosmo-api) running locally. Follow README to create tokens if you havent one yet and then add 
```
AI_API_AUTH_TOKEN = 'yourToken'
AI_API_URL = 'http://localhost:8000/api/classify/'
```
to your local.py 

Oh and the text on the tile is not yet adapted, but I thought this could wait until we have a new design for new tiles.


apps/classifications/signals.py
```diff
@@ -12,6 +12,7 @@
 
 client = httpx.Client()
 logger = logging.getLogger(__name__)
+ai_api_version = '3.0'
```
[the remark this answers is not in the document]
fuzzylogic2000: Yes, as we have the other API stuff like the url also in the settings, it would be good to add there. If you agree. :)

apps/classifications/signals.py
```diff
@@ -21,13 +22,29 @@ def get_ai_classification(sender, instance, **kwargs):
...
+                    if ai_api_version == '2.0' \
+                            and response.json()['classification'] == 'OFFENSE':
+
+                        classification = AIClassification(
+                            comment=instance,
+                            classifications=['OFFENSIVE']
+                        )
+                        classification.save()
+                    # version 3
+                    elif ai_api_version == '3.0' and \
```
fuzzylogic2000, line 36: we could also cast the string into a number and say 
```
if version <= 2:
    do this
else:
    do that
```
But as we don't really want to use versions <= 2.0, it's also fine like that and remove the whole if after the next release

fuzzylogic2000 (commented):
It reads very well! I will test it tomorrow.

apps/classifications/signals.py
Rineee: Yes, I think we can remove that soon, so left it like that!

phillimorland (approved):
Working! :) at least all 3 classifications are appearing, @Rineee I assume that's what what we anticipate right now? if so good to merge

Rineee:
> Working! :) at least all 3 classifications are appearing, @Rineee I assume that's what what we anticipate right now? if so good to merge

Cool! Yes, forgot to mention that, so far Anke didnt upload all of the different AIs, so they are all using the offensive model, thats why its either all categories or none for now :)

PR #176, opened 2022-04-06
settings/pyup: use fork-specific package file

fuzzylogic2000:
like https://github.com/liqd/diid_adplus/commit/35f0233c98b402f638ed03a1e69fc4f3780b8392


goapunk:
the ci files with ``ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/fork.txt'`` so I guess we need to create an empty file here?

fuzzylogic2000:
> the ci fails with `ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/fork.txt'` so I guess we need to create an empty file here for now or fill it with the fork dependencies?

Good catch, forgot to add it (git commit -a....). Now it should be fine @goapunk

goapunk:
nice!

goapunk (approved):

PR #178, opened 2022-04-07
#5762 userdashboard/ModerationStatement: adding dumb component for moderati…

phillimorland:
…on statment with styling


khamui:
this code is included in #173

PR #179, opened 2022-04-07
Pm 2022 04 issues

phillimorland:


khamui:
nice! about the fontawesome upgrades, we could prevent them by adding to only allow below v6 in renovate?

package.json
```diff
@@ -12,7 +12,7 @@
   ],
   "repository": "https://github.com/liqd/adhocracy-plus.git",
   "dependencies": {
-    "@fortawesome/fontawesome-free": "6.0.0",
+    "@fortawesome/fontawesome-free": "5.15.4",
```
fuzzylogic2000, line 15: This should have been done with the rebase. But if we just do a rebase soon, we can just skip this commit.

fuzzylogic2000:
> nice! about the fontawesome upgrades, we could prevent them by adding to only allow below v6 in renovate?

@khamui We shouldn't get renovate updates here as we do the upgrading in a+.

PR #180, opened 2022-04-10
Ks 2022 04 moderator statement

Rineee:
I cherry-picked the first commit from https://github.com/liqd/a4-kosmo/pull/172, so this closes https://github.com/liqd/a4-kosmo/pull/172.

I added a CommentDefaultRouter and CommentMixin the same way we do it for modules, organisations etc.. I added it here in contrib for now, so it is ready to use, but I think it would make sense to then move it to a4, where all the routers are, what do you think @fuzzylogic2000 ?  I would then make an issue or we can also do it now before rebase.

Usage is:
```
curl -X POST http://localhost:8004/api/comments/$comment_pk/moderatorstatement/ -H 'Authorization: Token $moderatorToken' -d 'statement=this is a statement'

curl -X PUT http://localhost:8004/api/comments/$comment_pk/moderatorstatement/$moderatorstatement_pk/ -H 'Authorization: Token $moderatorToken' -d 'statement=this is a statement UPDATED'

curl -X DELETE http://localhost:8004/api/comments/$comment_pk/moderatorstatement/$moderatorstatement_pk/ -H 'Authorization: Token $moderatorToken'

curl -X GET http://localhost:8004/api/comments/$comment_pk/moderatorstatement/ -H 'Authorization: Token $moderatorToken'

```


apps/moderatorfeedback/serializers.py
```diff
@@ -0,0 +1,28 @@
...
+        read_only_fields = ['last_edit', 'pk']
+
+    def create(self, validated_data):
+        validated_data['creator'] = self.context['request'].user
+        validated_data['comment'] = self.context['view'].comment
+
+        return super().create(validated_data)
+
+    def update(self, statement, validated_data):
+        validated_data['creator'] = self.context['request'].user
```
Rineee, line 21: This always sets creator to the last moderator that edited the statement.. not super sure if we want that?

apps/moderatorfeedback/models.py
```diff
@@ -52,3 +53,18 @@ class Moderateable(models.Model):
...
+        verbose_name=_('Moderator feedback'),
+    )
+    comment = models.OneToOneField(
+        Comment,
+        on_delete=models.CASCADE
+    )
+
+    @property
+    def project(self):
+        return self.comment.project
```
fuzzylogic2000, line 70: Could we also readd the string-method?
```
def __str__(self):
        return ('{} - {}').format(self.comment.id, self.statement)
```

fuzzylogic2000 (commented):
Perfect and ready to merge! (But I would like the string-method. Should I just add it?)

apps/moderatorfeedback/models.py
Rineee: Sure, just add it! Maybe then you could also add ListModelMixin to the API view, because I realized that for now, until we have the new serializer, we would need it?

Rineee: > Could we also readd the string-method?
> 
> ```
> def __str__(self):
>         return ('{} - {}').format(self.comment.id, self.statement)
> ```

Sorry, didnt delete that on purpose, just didnt see it..

fuzzylogic2000: Added both, but did not try it out again... Feel free to merge!

PR #181, opened 2022-04-12
apps/moderation/statement: editing statement behavior

khamui:


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -235,16 +236,18 @@ export default class ModerationNotification extends Component {
...
             initialStatement={this.state.moderatorStatement}
+            editing={this.state.isEditing}
           />}
-        {this.state.moderatorStatement &&
+        {this.state.moderatorStatement && !this.state.isEditing &&
           <ModerationStatement
             statement={this.state.moderatorStatement}
             onDelete={this.handleStatementDelete}
-            onEdit={this.handleStatementEdit}
+            onEdit={() => this.setState({ isEditing: true })}
```
Rineee, line 250: we could also set showModeratorStatementForm to true here and then only use that flag to show form?

Rineee, line 250: we could also set showModeratorStatementForm to true here and then only use that flag to show form?

Rineee:
Supercool! I just have one little suggestion, but if that doesnt work for you, we can also merge!

Rineee:
Supercool! I just have one little suggestion, but if that doesnt work for you, we can also merge!

khamui:
@Rineee good points there. I still need the `isEditing` for the textform component, so i left it there. Think its anyway better now.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -236,18 +236,18 @@ export default class ModerationNotification extends Component {
               <button className="btn btn--none" type="button" onClick={() => this.toggleIsPending()}><i className="fas fa-archive me-1" aria-hidden="true" />{unarchiveText}</button>
             </> /* eslint-disable-line react/jsx-closing-tag-location */}
         </div>
-        {(this.state.moderatorStatementForm || this.state.isEditing) &&
+        {this.state.showModeratorStatmentForm &&
           <ModerationStatementForm
             onSubmit={this.handleStatementSubmit}
             onEditSubmit={(payload, pk) => this.handleStatementEdit(payload, pk)}
             initialStatement={this.state.moderatorStatement}
             editing={this.state.isEditing}
```
khamui, line 244: i still need it here to distinguish between editing or adding new statement.

Rineee (approved):
Supercool, thanks for the changes!

PR #182, opened 2022-04-12
apps/userdashboard/notifications: making alerts appearing as design a…

khamui:
…nd with optional timer

### DIFF
--- a/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
+++ b/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
@@ -1,10 +1,16 @@
 import React, { Component } from 'react'
 import django from 'django'
 import api from './api'
-
 import { ModerationStatementForm } from './ModerationStatementForm'
-
 import { ModerationStatement } from './ModerationStatement'
+import Alert from '../../../../contrib/assets/Alert'
+
+const translated = {
+  statementAdded: 'Your note was successfully delivered.',
+  statementEdited: 'Your note was successfully updated.',
+  statementDeleted: 'Your note was successfully deleted.',
+  goToDiscussion: 'Go to discussion'
+}
 
 export default class ModerationNotification extends Component {
   constructor (props) {
@@ -15,7 +21,8 @@ export default class ModerationNotification extends Component {
       isBlocked: this.props.isBlocked,
       showModeratorStatmentForm: false,
       moderatorStatement: undefined,
-      isEditing: false
+      isEditing: false,
+      alert: undefined
     }
   }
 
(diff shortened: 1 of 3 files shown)


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -1,10 +1,16 @@
...
-
 import { ModerationStatement } from './ModerationStatement'
+import Alert from '../../../../contrib/assets/Alert'
+
+const translated = {
+  statementAdded: 'Your note was successfully delivered.',
+  statementEdited: 'Your note was successfully updated.',
+  statementDeleted: 'Your note was successfully deleted.',
+  goToDiscussion: 'Go to discussion'
+}
```
Rineee, line 14: translations! :) so we need ```django.pgettext('kosmo', 'string')``` here

Rineee, line 14: translations! :) so we need ```django.pgettext('kosmo', 'string')``` here

khamui: oops

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -82,7 +82,7 @@ export default class ModerationNotification extends Component {
           showModeratorStatementForm: false,
           alert: {
             type: 'success',
-            message: translated.statementAdded,
+            message: (() => { return (<>{translated.statementAdded} <a href={commentUrl}>{translated.goToDiscussion}</a></>) })(),
```
goapunk, line 85: @khamui I couldn't really come up with a nice solution without changing the alert component. Maybe that's too hacky?

khamui: hehe, well at least it works. I would keep the hacky solution and comment it in code?
later we can adjust the alert to make those kind of things (and maybe other things that is needed) work as well..? If you cool with that, feel free to merge @goapunk

khamui (commented):
nice! thanks for adding it

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
khamui: btw i think if you use `=> function` you can remove `{}` and `return`

goapunk: I turned it into a real function for now and added a comment

khamui:
cool! working fine

PR #184, opened 2022-04-13
translations: make po + some keys translated (notification mail)

khamui:


goapunk:
looks good to me, thanks!

PR #185, opened 2022-04-14
classifications/signals: only send comment to ai if created or text c…

Rineee:
…hanged


khamui:
nice! from reading code it looks good. not really tested, because i don't know how :)

PR #186, opened 2022-04-14
moderatorfeedback/rules: add view_moderatorcommentstatement

Rineee:


apps/moderatorfeedback/rules.py
```diff
@@ -16,3 +16,10 @@
     'a4_candy_moderatorfeedback.delete_moderatorcommentstatement',
     module_predicates.is_allowed_moderate_project
 )
+
+rules.add_perm(
+    'a4_candy_moderatorfeedback.view_moderatorcommentstatement',
+    module_predicates.is_allowed_moderate_project |
+    module_predicates.is_public_context |
```
goapunk, line 23: this needs to used so the new serializer will be able to get the for users?

goapunk (approved):
cool!

apps/moderatorfeedback/rules.py
Rineee: yes, that is so that all users can see the statements in the discussion.. I just copied the rule from view_comments :)

PR #187, opened 2022-04-14
styling: adding classes to style moderator statement on userfrontend

khamui:


fuzzylogic2000 (commented):
The code looks good, but I cannot see the moderator statement, also not after rebasing here in kosmo. Anything I am missing? Do I also have to rebase in a4?

goapunk:
@fuzzylogic2000 testing together with https://github.com/liqd/adhocracy4/pull/1045 works for me, it's collapsed by default though

fuzzylogic2000:
> @fuzzylogic2000 testing together with [liqd/adhocracy4#1045](https://github.com/liqd/adhocracy4/pull/1045) works for me, it's collapsed by default though

@goapunk But that's what I tried! With just updating the hash. Did you connect it locally? Also, doesn't matter. I probably did not clear my cache forceful enough or sth. like that. If it works for you, we can merge as soon as the PR in a4 is ready and merged.

goapunk:
Looks mostly good. The styling seems to be a bit off compared to the linked design though. e.g. the date has a 4px top margin but there shouldn't be any, instead it should have 8px at the bottom (or 6px, not sure) so the gap between moderator <-> date is smaller than between date <-> text

khamui:
also the last edit property has been adjusted in this PR.

mentioned in a4:
https://github.com/liqd/adhocracy4/pull/1045/commits/4a9fa7f1edd67bfc0505128d0b35780e868a9a05#r852974556

fuzzylogic2000 (commented):
Alright! Today it works. 
(@khamui @goapunk I only updated the hash in the python requirements yesterday. Obviously my holidays were too long. :p)
But there is a rogue "moderator" in front of the actual statement. And I guess, that is also why the statement is also shown when it's empty. I mean, this should be fixed in the a4 PR... And while you are at it, can you also add a space between the "Moderator" label of the creator of a comment and the date?
Screenshot from 2022-04-20 09-25-05
And then, you can also squash the commits. You don't need to put review changes into extra commits, but can just add them to the original one. I believe, that you changed it. :) Also, especially in the forks it is easier to rebase when there are as little commits as can be.
Apart from updating a4, that needs to be in an extra commit, because we want to skip it when rebasing.

goapunk:
There's also another small styling issue: the date has a different (darker) color in the design, here it's the same as the text

goapunk (approved):
looks great! thanks

fuzzylogic2000 (approved):
Nice! I will update a4 and merge!

fuzzylogic2000:
> Nice! I will update a4 and merge!

Ok, will not merge. @khamui Now the tests are failing.

PR #196, opened 2022-04-19
Ks 2022 04 add statement to comment serializer

Rineee:
https://github.com/liqd/adhocracy4/pull/1050 should be merged first and then the A4 hash updated here..


khamui:
working fine! I merged the corresponding a4 and updated this to use the latest main. Will test this further once its merged and connected to frontend components @Rineee .

apps/moderatorfeedback/serializers.py
```diff
@@ -26,3 +28,36 @@ def get_last_edit(self, statement):
...
+        model = Comment
+        read_only_fields = \
+            a4_serializers.CommentSerializer.Meta.read_only_fields + \
+            ('moderator_comment_statement',)
+        exclude = ('creator',)
+
+
+class CommentWithStatementListSerializer(CommentWithStatementSerializer):
+    """Serializer for the comments to be used when viewed as list."""
+
```
fuzzylogic2000, line 47: Does it work that empty? No `pass` or anything like that?

Rineee: I was also wondering, just copied that as it is from A4 and it seems to work, yes :woman_shrugging: 
https://github.com/liqd/adhocracy4/blob/84d56f6479d7072f15a4b06011d87f41027e8ca1/adhocracy4/comments_async/serializers.py#L202

fuzzylogic2000: Well, then. Ehm. Let's keep it like that? Or try if it still works when we do add the `pass`? Somehow looking at it makes me feel slightly uneasy. :rofl:

PR #198, opened 2022-04-20
Kt 2022 04 js tests

khamui:
covers components in userdashboard:  `<ModerationStatement>` and `<ModerationStatementForm>`


goapunk (commented):
Looks good to me! a sidenote: the ErrorList comes from aplus, right?  Maybe iit would make sense to add test there? We will then have it here and in all other projects after rebasing.

phillimorland:
@goapunk @khamui shouldn't we get the ErrorList from a4 as it's there too and so that's where test should be? i'm not sure how we did the js test set up with a4?

goapunk:
> @goapunk @khamui shouldn't we get the ErrorList from a4 as it's there too and so that's where test should be? i'm not sure how we did the js test set up with a4?

I didn't know it's there, if there is no good reason to have a separate copy of ErrorList in aplus I'm all for it

khamui:
good points people. will check and remove it here and in aplus if possible

goapunk:
cool, coverage increased even more now ;-)

khamui:
not sure if this matters for now but on local a4 component files are loaded from:
`venv/lib/python3.7/site-packages/adhocracy4/`

and on the server the files are being loaded from:
`static/`

 @fuzzylogic2000 or @Rineee if you have an idea why it is like that?

phillimorland (approved):
very nice :) will leave the site packages in for now, we can adjust if needed

PR #199, opened 2022-04-21
Ks 2022 04 comment notification statement serializer

Rineee:
Left the old apis as they are for now as we might still want to use them.

The viewset is available under /api/userdashboard/moderation/$project_pk/comments/ and I added routes to archive/unarchive,
they are /api/userdashboard/moderation/$project_pk/comments/$comment_pk/archive and /unarchive, sending GET to these will automatically archive/unarchive all notifications of the respective comment.

To Do:

- [ ] - integrate moderator statement api (maybe just serialize url and leave api as it is?)
- [x] - filtering, this includes annotations I think
- [x] - sorting, also requires annotations (this is an own story, see https://taiga.liqd.net/project/mkay-kosmo/us/114?no-milestone=1)
- [x] - connect to frontend and remove old stuff
- [ ] - maybe routes to view classifications as we already have the serializers/apis?
- [ ] - permissions


apps/userdashboard/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+    class Meta:
+        model = Comment
+        fields = ['category_counts', 'comment', 'comment_url',
+                  'last_edit', 'pk', 'has_pending_notifications', 'is_blocked',
+                  'is_moderator_marked', 'moderator_statement',
+                  'time_of_last_notification', 'user_image', 'user_name',
+                  'user_profile_url']
+
+    # helper
+    def has_pending_and_archived_notifications(self, comment):
```
fuzzylogic2000, line 36: I think if we only used internally, there is the convention to call it `_has_pending_and_archived_notifications`.

fuzzylogic2000: Same for the other helpers.

fuzzylogic2000 (commented):
Supernice! I couldn't find anything wrong with the API, only a naming convention thingy.

apps/userdashboard/serializers.py
Rineee: Oh yes, good point, will change that!

phillimorland (approved):
reads very nicely! and it works :)

fuzzylogic2000 (commented):
Supercool!

PR #201, opened 2022-04-21
Using Alert from a4

khamui:


fuzzylogic2000 (commented):
Does look good! But the second commit needs to be done in a+ instead.

apps/documents/assets/DocumentManagement.jsx
```diff
@@ -1,11 +1,11 @@
+import { alert as Alert } from 'adhocracy4'
 const api = require('adhocracy4').api
 const React = require('react')
 const django = require('django')
 const dashboard = require('adhocracy4/adhocracy4/dashboard/assets/dashboard')
 const update = require('immutability-helper')
 const ChapterNav = require('./ChapterNav')
 const ChapterForm = require('./ChapterForm')
-const Alert = require('../../contrib/assets/Alert')
```
goapunk, line 8: this should be done in aplus, we'll get the change after rebasing :)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationStatementForm.jsx
```diff
@@ -24,15 +22,8 @@ export const ModerationStatementForm = (props) => {
...
 
-  const removeAlert = (e) => {
-    this.setState({
-      alert: null
-    })
-  }
-
   return (
     <form className="general-form" onSubmit={handleSubmit}>
-      <Alert onClick={() => removeAlert()}>{translated.alertError}</Alert>
```
goapunk, line 35: sorry, not fimiliar with the code: we remove this because it was unused? the translation for `translated.alertError` can be removed then as well, right?

khamui: oh yes..

Issue #202, opened 2022-04-25
moderation statement and filtering update does not work


Rineee:
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** when I look at filtered list of archived comments and click on unarchive on one of the comments that has a moderator statement, the comment disappears, but statement remains and is shown under the comment above
**important screensize:**
**device & browser:** 
**Comment/Question:** same for pending

Screenshot?

PR #206, opened 2022-04-27
Ks 2022 04 connect moderation statement api

Rineee:
Fixes https://github.com/liqd/a4-kosmo/issues/202

Also makes statement not editable once archived.

### DIFF
--- a/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
+++ b/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
@@ -74,16 +74,6 @@ export default class ModerationNotificationList extends Component {
     this.setState({ alert: undefined })
   }
 
-  filterNotifications = () => {
-    return this.state.moderationComments.filter(comment => {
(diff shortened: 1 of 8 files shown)


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
```diff
@@ -74,16 +74,6 @@ export default class ModerationNotificationList extends Component {
...
-  filterNotifications = () => {
-    return this.state.moderationComments.filter(comment => {
-      return this.state.filterItem.filter === ''
-        ? true
-        : this.state.filterItem.name === 'Pending'
-          ? comment.has_pending_notifications
-          : !comment.has_pending_notifications
-    })
-  }
-
```
Rineee, line 86: @kha I removed this, because I found that it is not really needed as the filtering is already done when fetching.. Or did I miss something?

khamui (approved):
lgtm. I would test it further when working on the ModerationNotification in a bit.
We could merge, except @phillimorland has any objections, because there are some improvements coming from her as well.

PR #207, opened 2022-04-27
notification tests

fuzzylogic2000:
In the test email command, the delete project email is broken. But that has to be fixed in a+. To test, just uncomment the broken email.

### DIFF
--- a/tests/notifications/test_creator_notifications.py
+++ b/tests/notifications/test_creator_notifications.py
@@ -87,3 +87,85 @@ def test_notify_creator_on_moderator_feedback(proposal_factory, client):
     assert mail.outbox[1].to[0] == creator.email
     assert mail.outbox[1].subject.startswith(
         'Feedback for your contribution')
+
+
+@pytest.mark.django_db
+def test_notify_comment_creator_on_block(
+        idea, comment_factory, apiclient, admin,
+        ai_classification_factory):
+    """Check if creator gets email on comment blocked."""
+    comment = comment_factory(content_object=idea)
+    ai_classification_factory(comment=comment)
+    creator = comment.creator
+    # FIXME: use moderator instead of admin when rules are there
+    # moderator = comment.project.moderators.first()
+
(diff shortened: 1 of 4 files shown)


tests/notifications/test_creator_notifications.py
```diff
@@ -87,3 +87,53 @@ def test_notify_creator_on_moderator_feedback(proposal_factory, client):
...
+            'project_pk': comment.project.pk,
+            'pk': comment.pk,
+        }
+    )
+    data = {
+        "is_blocked": True,
+    }
+
+    # apiclient.force_authenticate(user=admin)
+    # response = apiclient.patch(url, data, format='json')
```
fuzzylogic2000, line 116: Here, I get a 404, and I don't get that! @Rineee do you have an idea? But anyway, that is a problem for Monday.

fuzzylogic2000: Ah! Thank you! That makes total sense, also the 404 then. :) I will try that!

tests/notifications/test_creator_notifications.py
```diff
@@ -87,3 +87,85 @@ def test_notify_creator_on_moderator_feedback(proposal_factory, client):
...
+@pytest.mark.django_db
+def test_notify_comment_creator_on_block(
+        idea, comment_factory, apiclient, admin,
+        ai_classification_factory):
+    """Check if creator gets email on comment blocked."""
+    comment = comment_factory(content_object=idea)
+    ai_classification_factory(comment=comment)
+    creator = comment.creator
+    # FIXME: use moderator instead of admin when rules are there
+    # moderator = comment.project.moderators.first()
```
fuzzylogic2000, line 101: @Rineee If we merge your's first, I can do this. Otherwise, could you add the fix to your PR?

Rineee: I'll do it in mine!

Rineee (approved):
Very nice!

PR #209, opened 2022-04-28
WIP Pm 2022 04 js tests

phillimorland:


apps/userdashboard/assets/js/a4_candy_userdashboard/__tests__/ModerationStatement.jest.jsx
```diff
@@ -20,13 +20,31 @@ test('ModerationStatement onDelete', () => {
...
-  render(
+  const tree = render(
     <ModerationStatement
       statement={mockProps}
       onDelete={mockOnDelete}
       notificationIsPending
     />)
-  const deleteButton = document.querySelector('li button')
+  const deleteButton = tree.container.querySelector('#delete-input')
   fireEvent.click(deleteButton)
```
phillimorland, line 30: @khamui @goapunk 
We do this in our testing in a4 and thought it made it more precise then just calling document, what do you guys think? 
Also I wasn't sure how else to target different buttons so I just added an id, there is a test id thing but I got confused but if we think this is too hacky can look into it, or we just make a general test for any button on statement which I think is what it was doing before?!

fuzzylogic2000 (approved):
Cool!

PR #210, opened 2022-04-28
[#5764] apps/userdashboard: utilize moderator_marked field for highlight

sabinammm:
To test locally

- Get token using 

`curl -X POST http://localhost:8004/api/login/ -H 'Accept: application/json;' -d 'username=<username>&password=<password>'`
- Add token, project_pk and comment_pk to 

`curl -X PATCH http://localhost:8004/api/userdashboard/moderation/<project_pk>/comments/<comment_pk>/ -H 'Accept: application/json;' -H 'Authorization: Token <token>' -d 'is_moderator_marked=True'`

### DIFF
--- a/apps/userdashboard/api.py
+++ b/apps/userdashboard/api.py
@@ -82,6 +82,10 @@ class ModerationCommentViewSet(mixins.ListModelMixin,
             comment.save()
             if comment.is_blocked:
                 NotifyCreatorOnModeratorBlocked.send(comment)
+        elif 'is_moderator_marked' in self.request.data:
+            comment = self.get_object()
+            comment.is_moderator_marked = request.data['is_moderator_marked']
+            comment.save()
         return super().update(request, *args, **kwargs)
 
     @action(detail=True)


Rineee:
Yes, that would be right! Buuut I just realized that this is not needed anymore.. I am super sorry for extra work, I copied that over from the old api and forgot to delete it. The old api was on the cĺassification model, so, that is why we needed to do this in order to change the comment model. But since this new api is on the comments directly, that is not needed anymore, the update just works automatically.. really sorry, should have deleted that before.
But we still need the part with the e-mail.. so: if you want, you could remove the whole part about updating the comment, and only check if 'is_blocked' is in request.data and if it is set to true send the e-mail.

Rineee (changes requested):
Im sorry, that would work, but unfortunately not needed anymore..

khamui:
@Rineee but its about the `Highlight` no? For this I think we do not have an Email, yet or do we?

Linking the frontend part:
https://github.com/liqd/a4-kosmo/pull/211#discussion_r861010730

khamui:
Ah got it! makes sense

fuzzylogic2000:
Sorry! I didn't notice and put you on the wrong (or unneeded) track there.

@Rineee should I do the blocked email-thingy along with the tests now?

PR #211, opened 2022-04-28
icons/userdashboard: adding highlight icon

khamui:


adhocracy-plus/assets/scss/icons/style.css
```diff
@@ -75,3 +75,52 @@
 .font--lg {
```
khamui, line 75: this i just saw while solving a cherry-pick conflict, but this seems to be wrong here..

khamui:
merging, as building was successful

PR #213, opened 2022-05-02
DISCUSS: buttons/outline: show blue outline when using tab to navigate (note: …

khamui:
…will show blue outline on mouse click as well)

Not sure if we want this, with this tabbing shows the blue outline, but on simple click it will show it as well. A11y in terms of semantic is given anyways because buttons.

### DIFF
--- a/adhocracy-plus/assets/scss/components/_button.scss
+++ b/adhocracy-plus/assets/scss/components/_button.scss
@@ -75,6 +75,11 @@
 .btn {
     @extend %button-base;
     @include button-color($link-color, $link-hover-color);
+
+    &:focus {
+        outline: 2px solid Highlight !important;
+        outline: 5px auto -webkit-focus-ring-color !important;
+    }
 }
 
 .btn--light {


fuzzylogic2000 (commented):
Shouldn't a focus ring be there already with the button-color? https://github.com/liqd/a4-kosmo/blob/a4b627df28f1ea85de994cbe5ddfcb192ecc598c/adhocracy-plus/assets/scss/components/_button.scss#L10
Do we overwrite that somewhere for the MD buttons?
Ah, I think we sometimes do not use that mixin. Hmmm. I think, we should just not whack it at the button with `important`, though. Or should we @phillimorland ? :woman_shrugging: 
Maybe let's keep it like this and discus and fix after discussion.

khamui:
ah, I think that was what i didn't get from reading code! i was searching where it is defined (didn't know the mixin thingy).
At the end I just copied the lines from `_datepicker.scss`. Yes lets see what the discussion brings and then fix..

phillimorland:
@khamui @fuzzylogic2000 for default btn styling it should probably go into a+ right? and if we want this behaviour it should be in the btn mixin so we should not add it with important. I'm also not sure why we need this behaviour as outline with tabbing works? and it we do want to implement it we should get a design as it will be everywhere and a constant part of the user interaction.

khamui:
@phillimorland for me it doesn't work (i mean the tabbing with outline does not appear on buttons, just on links). I don't know if this is a requirement, I thought it is a a11y requirement, but if not we can leave as is.

tested in Chrome.

phillimorland:
@khamui the tabbing was fixed in a+ last week, you merged it :p https://github.com/liqd/adhocracy-plus/pull/1812 it will be here after rebase

khamui:
ohh...right! okay sorry i forgot about this dependency. well then all fine

PR #216, opened 2022-05-02
apps/userdashboard/ModerationStatement: Fixing refreshing

khamui:


khamui:
The last commits is necessary, since we are fetching the comment list now it could lead to a bit of waiting time (in this case we show loading, to make sure users do not take another action (blocking, highlithing, deleting etc.))

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -292,16 +305,22 @@ export const ModerationNotification = (props) => {
...
-            onToggleForm={() => toggleModerationStatementForm()}
-            onToggleBlock={() => toggleIsBlocked()}
-            onToggleHighlight={() => toggleIsHighlighted()}
-            onTogglePending={() => toggleIsPending()}
-          />
+          {loading
+            ? (
+              <div className="d-flex justify-content-center">
+                <i className="fa fa-spinner fa-pulse" aria-hidden="true" />
+              </div>)
```
phillimorland, line 312: this shrinks the notification and looks broken when loading is fast, would add some height somehow so it doesn't look like the notification collapses

phillimorland (commented):
Looking good, was the change from class to function to do with the life cycle? maybe you can add an explanation in PR description so if we ever need to see why it's recoded somewhere. Also does this mean we should re-assess our current policy with class and functional components?

khamui:
> Looking good, was the change from class to function to do with the life cycle? maybe you can add an explanation in PR description so if we ever need to see why it's recoded somewhere. Also does this mean we should re-assess our current policy with class and functional components?

Technically it should work also for class notation. Actually I tried it first to refactor with the class version, but with the handling of `this.state` it caused a lot of breakages for me. Turning it into function notation helped me to understand the problem.

So generally I prefer the function notation, what do you think @phillimorland ?

phillimorland:
> > Looking good, was the change from class to function to do with the life cycle? maybe you can add an explanation in PR description so if we ever need to see why it's recoded somewhere. Also does this mean we should re-assess our current policy with class and functional components?
> 
> Technically it should work also for class notation. Actually I tried it first to refactor with the class version, but with the handling of `this.state` it caused a lot of breakages for me. Turning it into function notation helped me to understand the problem.
> 
> So generally I prefer the function notation, what do you think @phillimorland ?

I think we should try to be as consistent as possible, this seems that we are breaking from our decision to have class for stateful components and functional for the rest. If this is how this has to be done then there should be an explanation, either in comments or in the PR.
As for changing the policy, it feels like we only just implemented some kind of default react rules for our code and I would like to avoid changing them with every new feature so that should be a bigger discussion

phillimorland (approved):

PR #217, opened 2022-05-03
apps/userdashboard/ModerationNotification/HoverButton: adding compone…

khamui:
…nt to handle hover text changes

squash


apps/userdashboard/assets/js/a4_candy_userdashboard/HoverButton.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+  return (
+    <button
+      id={props.id}
+      className={props.className}
+      type="button"
+      onClick={props.onClick}
+      disabled={props.disabled}
+      onMouseEnter={() => setButtonText(props.children.hoverText)}
+      onMouseLeave={() => setButtonText(props.children.text)}
```
phillimorland, line 14: is there a reason not to just use a title attribute here?

apps/userdashboard/assets/js/a4_candy_userdashboard/HoverButton.jsx
```diff
@@ -0,0 +1,22 @@
+import React, { useState } from 'react'
```
phillimorland, line 1: I'm wondering as this is a general component if it would be better in the contrib folder?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationActionsBar.jsx
```diff
@@ -1,15 +1,19 @@
...
 export const ModerationNotificationActionsBar = (props) => {
   const translated = {
-    blockText: django.pgettext('kosmo', ' Block'),
-    unblockText: django.pgettext('kosmo', ' Unblock'),
-    replyText: django.pgettext('kosmo', ' Add statement'),
-    unarchiveText: django.pgettext('kosmo', ' Unarchive'),
+    blockText: django.pgettext('kosmo', 'Block'),
+    unblockText: django.pgettext('kosmo', 'Unblock'),
+    isBlockedText: django.pgettext('kosmo', 'Blocked'),
+    replyText: django.pgettext('kosmo', 'Add statement'),
```
phillimorland, line 10: this is now too close to icon

apps/userdashboard/assets/js/a4_candy_userdashboard/HoverButton.jsx
khamui: i am not following i think. Do you mean adding attributes for title, hovertext and icon? yes that would be possible, thought its more natural to place the children between the tags.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationActionsBar.jsx
phillimorland: check it, looked wrong on my browser

apps/userdashboard/assets/js/a4_candy_userdashboard/HoverButton.jsx
phillimorland: you shouldn't need the mouseEnter or mouseLeave stuff at all unless i'm missing something? https://www.w3schools.com/tags/att_global_title.asp

khamui: could you please make an example how to use this for the hover effect? i am not following...sorry

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationActionsBar.jsx
khamui: okay i will triple check, for me there is a space. Anyway should we include spacing in a translation key? that could lead to redundancy no? I think we should separate text from spacing

khamui: okay you were right, the space didn't appear. Using a span now, hope that works

apps/userdashboard/assets/js/a4_candy_userdashboard/HoverButton.jsx
phillimorland: is the link not working? it's explains the attribute and shows an example. If all we need is a tooltip on hover then we should be able to use that attribute, if we need additional functionality then maybe not? I am unsure as there isn't a US for it?!

phillimorland: ahhh now I get it, cheers! ok not sure that was working when I checked, will look at it again

phillimorland (changes requested):
the on hover behaviour works but when I click `highlight` move mouse away, the button still says highlight until I move it back onto the button, then it says `unhighlight` then mouse moves out and it says `highlighted`, my understanding is it should say `highlighted` straight after the notification is reloaded, not after additional hover. Also don't think unhighlight is a word but not sure what to do about that?

khamui:
this continues living in another PR #221 (rebased version)

PR #221, opened 2022-05-04
apps/userdashboard/ModerationNotification/HoverButton: adding compone…

khamui:
…nt to handle hover text changes

squash

squash

### DIFF
--- a/apps/contrib/assets/HoverButton.jsx
+++ b/apps/contrib/assets/HoverButton.jsx
@@ -1,0 +1,27 @@
+import React, { useState, useEffect } from 'react'
+
+export const HoverButton = (props) => {
+  const [buttonText, setButtonText] = useState(props.text)
+
+  useEffect(() => {
+    setButtonText(props.text)
+  }, [props.text])
+
+  return (
+    <button
+      id={props.id}
+      className={props.className}
+      type="button"
+      onClick={props.onClick}
+      disabled={props.disabled}
+      onMouseEnter={() => setButtonText(props.hoverText)}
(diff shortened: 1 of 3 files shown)


khamui:
okay will keep working on this one:
-[ ] issue with updating word after click
-[ ] `unhighlight` is not a word, alternatives?

khamui:
@phillimorland indeed it didn't update after taking action (highlight, or blocked). That should be fixed now. Anyways there is a tiny little delay though. Had a hard time to find a solution, but couldn't manage to find one. Let me know if this is too annoying and I should try find a way to solve it or if I should make an issue and we go with this for now?

phillimorland (approved):
I think delay is fine, we can user test it if UX or PM's think it's an issue, nicely done! :)

goapunk:
@phillimorland @khamui I'm a bit torn if we should handle it like this because it's rather working by accident, not by design. Maybe we can come up with a more robust solution? The problem here is that the `onMouseEnter` event isn't "reset" after clicking the button if you don't move the cursor away from it, so the hover effect is not triggered again. Showing the loading icon is basically a hack/trick to force that but looks rather ugly imo. One alternative for example could be to have a variable in the HoverButton which saves whether the cursor is on it, and set the text depending on the current value of that. This way we don't need to re-render the whole `notificationActionbar` but really just change the text of the button.

phillimorland:
@khamui @goapunk ahh totally missed that, yes we should definitely make it as robust as possible! who wants to take that?

khamui:
alright, thanks for the feedbsck guys! I am also not happy with the solution. will check how twitter solves it and mimic that

PR #222, opened 2022-05-04
milestone issues fixes 

### LINKED ISSUE #191: #5763 Display 2 more categories of comments reported by AI
**URL:** https://kosmo-dev.liqd.net/userdashboard/moderation/detail/rines-testproject/
**user:** *moderator*
**important screensize:** *Both*
**device & browser:** *Safari Version 15.3 (17612.4.9.1.8) & Chrome Version 99.0.4844.51*
**Comment/Question:** *Seems fine, I created 4 colours in case they want to differentiate them. They are #EBE8FF  #FCE5E5 #E7E7F1 #FCE5EC* 

Screenshot 2022-04-19 at 10 01 49

### LINKED ISSUE #192: #5762 Moderator can edit moderator remark
**URL:** https://kosmo-dev.liqd.net/userdashboard/moderation/detail/rines-testproject/
**user:** *moderator*
**important screensize:** *Both*
**device & browser:** *Safari Version 15.3 (17612.4.9.1.8) & Chrome Version 99.0.4844.51*
**Comment/Question:** *If possible, instead of disabling the button, shall we change the Add Statement button to Edit Statement once the Statement is added?* 

Screenshot 2022-04-19 at 10 17 45

khamui:
closes #192 
closes #191

### DIFF
--- a/adhocracy-plus/assets/scss/_variables.scss
+++ b/adhocracy-plus/assets/scss/_variables.scss
@@ -40,7 +40,11 @@ $text-muted:    $text-color-gray !default;
 
 $blue-light: #d6eff9;
 $purple-light: #ecdbff;
+$purple-lighter: #ebe8ff; // classification: engaging
+$purple-lightest: #e7e7f1; // classification: ai
 $green-light: #e0fbe2;
+$red-light: #fce5e5; // classifiction: offensive
+$red-lighter: #fce5ec; // classifiction: fact claiming
 
 $print-black: #000 !default;
 $print-link: #005cb4 !default;
--- a/adhocracy-plus/assets/scss/components/_a4-comments.scss
+++ b/adhocracy-plus/assets/scss/components/_a4-comments.scss
@@ -272,6 +272,22 @@
     background-color: $green-light;
 }
 
+.a4-comments__badge[data-classification="offensive"] {
+    background-color: $red-light;
+}
+
+.a4-comments__badge[data-classification="engaging"] {
+    background-color: $purple-lighter;
+}
+
+.a4-comments__badge[data-classification="fact claiming"] {
+    background-color: $red-lighter;
+}
+
+.a4-comments__badge[data-classification="ai"] {
+    background-color: $purple-lightest;
+}
+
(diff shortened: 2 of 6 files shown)


fuzzylogic2000:
The code looks good to me, but the tests are failing.

khamui:
passing now :) @fuzzylogic2000

fuzzylogic2000 (approved):
Looks good! 
But should we also remove the edit from the dropdown then? That is probably a question for @mcastro-lqd though. (See screenshot)
Screenshot from 2022-05-05 12-08-58
If the answer is no, we can merge. :)

phillimorland:
@khamui also sorry there are now conflicts :(

adhocracy-plus/assets/scss/_variables.scss
```diff
@@ -138,3 +138,9 @@ $github: #333;
 $utilities-text-colors: #333; // <-- BS inport fix
 $utilities-colors: (); // <-- BS inport fix
 $utilities-text-colors: (); // <-- BS inport fix
+
+// classifications
+$offensive: #FCE5E5 !default;
+$engaging: #EBE8FF !default;
+$factclaiming: #FCE5EC !default;
+$ai: #E7E7F1 !default;
```
phillimorland, line 146: hexs should all be lowercase not sure why the linter didn't catch that? also in this case I would use colour names as we may want them elsewhere and the grey we can update an existing variable, same with the purple as we already have a `$purple-light` we don't want loads of different colours if we can possible help it!

khamui: out of curiosity, i used `!default` but actually without knowing what it does. Do you have an explanation what it does. couldn't find any useful information about it @phillimorland ?

phillimorland: it just sets a default for that variable name, we load our `variables` file before we load the bootstrap one in style.scss so the defaults we want are set and the bootstrap ones get ignored because there is already one set.

phillimorland (approved):
cool

PR #223, opened 2022-05-05
[5855] apps/userdashboard: show time of last notification in moderato…

fuzzylogic2000:
…r dachboard


phillimorland (approved):
looking fab, just realised the date format is different from designs but it fits with comments so I think we just wait and see what they say

PR #224, opened 2022-05-06
Ks 2022 05 classifications filter

Rineee:
Probably needs more styling? I added a task..

### DIFF
--- a/apps/userdashboard/api.py
+++ b/apps/userdashboard/api.py
@@ -1,7 +1,10 @@
+from django.db.models import Case
 from django.db.models import Exists
 from django.db.models import ExpressionWrapper
 from django.db.models import OuterRef
 from django.db.models import Q
+from django.db.models import Value
+from django.db.models import When
 from django.db.models.fields import BooleanField
 from django.shortcuts import get_object_or_404
 from django_filters.rest_framework import BooleanFilter
@@ -10,6 +13,7 @@ from django_filters.rest_framework import FilterSet
 from rest_framework import mixins
 from rest_framework import viewsets
 from rest_framework.decorators import action
+from rest_framework.filters import BaseFilterBackend
 from rest_framework.response import Response
 
 from adhocracy4.api.permissions import ViewSetRulesPermission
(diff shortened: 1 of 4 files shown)


apps/userdashboard/api.py
```diff
@@ -25,6 +29,32 @@
...
+                    When(has_pending_notifications=True, then=Value(True)),
+                    When(has_pending_notifications=False, then=Value(False))
+                ),
+                  ai_classifications__classifications__contains=classification) |
+                Q(user_classifications__is_pending=Case(
+                    When(has_pending_notifications=True, then=Value(True)),
+                    When(has_pending_notifications=False, then=Value(False))
+                ),
+                  user_classifications__classifications__contains=classification)
+            )
```
fuzzylogic2000, line 54: Wow, complicated and nice!

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
```diff
@@ -2,42 +2,73 @@ import React, { Component } from 'react'
...
-      filterItem: filterItem,
+      selectedFilters: {
+        ...this.state.selectedFilters,
+        pending: value
+      },
       isLoaded: false
-    })
-    this.loadData(filterItem.filter)
+    },
+    () => this.loadData()
```
fuzzylogic2000, line 49: Sorry, missing React-knowledge: could you explain what the `...` and `() =>` does?

fuzzylogic2000 (approved):
Supercool!

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
Rineee: @fuzzylogic2000 Sure! The ... is the spread operator, same as ** in python. It unpacks the key/values of the dictionary, so here it means, that it takes all entries from this.state.selectedFilters.

The () => is function notation, I needed that, because I use loadData() here as a callback to setState (so it only loads the date once the state is updated), which expects a function. But since loadData itself is async, it returns a promise and thus I needed to wrap it again in a function (also found that by try and error).

Hope that makes sense?

PR #227, opened 2022-05-09
apps/userdashboard/moderationNotification/ActionsBar/HoverButton: ref…

khamui:
…actor to show hoverbutton instantly (without loading).

To mention: the disabling is still a little bit delayed because this is set by a variable which has to come from the `list re-fetching`.

### DIFF
--- a/apps/contrib/assets/HoverButton.jsx
+++ b/apps/contrib/assets/HoverButton.jsx
@@ -1,21 +1,31 @@
 import React, { useState, useEffect } from 'react'
 
 export const HoverButton = (props) => {
-  const [buttonText, setButtonText] = useState(props.text)
+  const { textMouseOn, textMouseOff, onClick } = props
+  const [buttonText, setButtonText] = useState(textMouseOff)
+  const [processing, setProcessing] = useState(false)
+
+  const handleClick = () => {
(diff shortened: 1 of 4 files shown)


khamui:
In order to test "hover" through tabbing, you might want to set the style `outline` on focus by hand (this comes from a rebase of aplus).

khamui:
@phillimorland @goapunk after thinking again about this i decided to remove a huge complexity (this criteria stuff), which i shifted into the HoverButton component before. But it is definitely not the responsiblity of the button to deal with any criterias.

So instantly changing the button state is not possible by design, because we fetch this information nested in a parent list. I think twitter fetches the `follow` state immediately (inside of the component) on an endpoint called `/friendships`, but not within a parent list. That is why they can set it instantly. For us this would be a hack (with this complexity we had before).

long story short: I think in order to have clean code we have to deal with the time the list gets fetched again. Currently i am `disabling` the button while fetching.

phillimorland (approved):
yo yo much cleaner code, nice one, the functionality might need a little tweeking but I think thats a choice for Marcelo and we won't know the actual load speed unless we try in on a live site so think we should merge and see :)

PR #230, opened 2022-05-11
Ks 2022 05 issues

Rineee:


apps/userdashboard/serializers.py
```diff
@@ -174,3 +176,29 @@ def get_num_active_notifications(self, comment):
             return self._get_num_pending_notifications(comment)
         else:
             return self._get_num_all_notifications(comment)
+
+    def update(self, instance, validated_data):
```
Rineee, line 180: This is essentially the update method from rest_framework.serializers.ModelSerializer, I just added ignore_modified=True to the instance.save()

fuzzylogic2000: Could you add a link to that method in a docstring? Not sure that is the best way to remember what we did, but might help.

fuzzylogic2000 (approved):
Nice! Just maybe add some more explanation for the first commit in the code?

apps/userdashboard/serializers.py
Rineee: Yes, good idea!

fuzzylogic2000 (approved):
Whoop whoop!

PR #235, opened 2022-05-17
Pm 2022 05 release issues

phillimorland:


adhocracy-plus/assets/scss/components/_a4-comments.scss
```diff
@@ -28,6 +28,7 @@
 
     mark > p {
         background-color: $purple-light;
+        display: inherit;
```
khamui, line 31: brilliant!

khamui:
very nice!

PR #242, opened 2022-05-19
Ks 2022 05 classifications export

Rineee:
depends on https://github.com/liqd/adhocracy4/pull/1084, so needs update once that is merged


fuzzylogic2000 (approved):
Very nice!

PR #250, opened 2022-05-23
WIP apps/userdashboard/ModerationNotification.jsx: making german classifi…

khamui:
…cations with color.

Better would be to receive a `type` or something on the endpoint, for instance:

```
category_counts: [
  { type: 'offensive', text: 'beleidigend', count: 1 },
  ...
]
```

so i do not have to rely on translated words. Would this be possible?

### DIFF
--- a/adhocracy-plus/assets/scss/components/_a4-comments.scss
+++ b/adhocracy-plus/assets/scss/components/_a4-comments.scss
@@ -275,15 +275,18 @@
     background-color: $green-light;
 }
 
-.a4-comments__badge[data-classification="offensive"] {
+.a4-comments__badge[data-classification="offensive"],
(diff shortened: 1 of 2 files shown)


Rineee:
> @khamui sending a type is fairly easy I think. Does it need to be a string though? Or would it be enought to just send an Int (less data usage) ?



> …cations with color.
> 
> Better would be to receive a `type` or something on the endpoint, for instance:
> 
> ```
> category_counts: [
>   { type: 'offensive', text: 'beleidigend', count: 1 },
>   ...
> ]
> ```
> 
> so i do not have to rely on translated words. Would this be possible?

Yes! I will check on how to do it best!

khamui:
Think this is being fixed with the bigger PR of @Rineee

PR #258, opened 2022-05-31
notifications/emails/notify_creator_on_moderator_blocked: cta leading…

khamui:
… to discussions and take full comment instead of partly

### DIFF
--- a/apps/notifications/templates/a4_candy_notifications/emails/notify_creator_on_moderator_blocked.en.email
+++ b/apps/notifications/templates/a4_candy_notifications/emails/notify_creator_on_moderator_blocked.en.email
@@ -34,11 +34,13 @@
 </strong>
 
 <p style="margin-bottom: 1em">
-    {{ comment }}
+    {{ comment.comment }}
 </p>
 {% endblock %}
 
-{% block cta_url %}{% if comment.content_object.get_absolute_url %}{{ email.get_host }}{{ comment.content_object.get_absolute_url }}{% else %}{{ email.get_host }}{{ module.get_detail_url }}{% endif %}{% endblock %}
+{% block cta_url %}
+{{ email.get_host }}{{ discussion_url }}
+{% endblock %}
 
 {% block cta_label %}
 {# Translators: KOSMO #}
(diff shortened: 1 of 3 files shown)


apps/notifications/templates/a4_candy_notifications/emails/notify_creator_on_moderator_blocked.en.email
```diff
@@ -34,11 +34,21 @@
 </strong>
 
 <p style="margin-bottom: 1em">
-    {{ comment }}
+    {{ comment.comment }}
```
khamui, line 37: small fix: we want to show the full comment

apps/notifications/templates/a4_candy_notifications/emails/notify_creator_on_moderator_blocked.en.email
```diff
@@ -34,11 +34,21 @@
...
+{% block cta_url %}
+{% if comment.parent_comment.exists %}
+    {{ email.get_host }}{{ comment.content_object.content_object.get_absolute_url }}
+{% else %}
+    {% if comment.content_object.get_absolute_url %}
+        {{ email.get_host }}{{ comment.content_object.get_absolute_url }}
+    {% else %}
+        {{ email.get_host }}{{ module.get_detail_url }}
+    {% endif %}
+{% endif %}
```
fuzzylogic2000, line 50: Nice! But can we put that logic into the email.py? https://github.com/liqd/a4-kosmo/blob/7302118e141872cfc9de2e3449816266a387b735/apps/notifications/emails.py#L107
Templates should be rather thin on logic, even though we do not always follow that rule.

fuzzylogic2000: I mean, you can leave the get_host on here (we only do that for the netiquette because it has to go into the translated string), but put the url into the context and then you only have to do `{{ email.get_host }}{{ discussion_link }}` in the template without all the ifs and elses.

fuzzylogic2000: Oh, and it can go into an if/elif/else instead.

fuzzylogic2000:
@khamui Done! You can merge when it's also fine with your comments, you tried it on. But should be as I used your exact if/else, but only put it elsewhere.

PR #259, opened 2022-05-31
userdashboard/api: use moderation_comment rules for OPTIONS request

Rineee:
I forgot to overwrite permissions for OPTIONS requests, so it was throwing an error when not logged in as Admin and trying to access the userdashboard api, eg kosmo-dev.liqd.net/api/userdashboard/moderation/1/comments/49/userclassifications/ (Attention, this throws a 500 :)) So this fixes it.


fuzzylogic2000 (approved):
Ah! I saw the sentry. :) Nice!

PR #291, opened 2022-08-05
Ks 2022 08 platform information

Rineee:
Not sure about the styling, but I tried :)


adhocracy-plus/assets/scss/components/_userdashboard.scss
```diff
@@ -9,6 +9,10 @@
     padding-bottom: 2 * $padding;
 }
 
+.userdashboard-profile__manual {
+    border: 1px solid #dadada;
```
fuzzylogic2000, line 13: It does look great and right!
But I would rather add some kind of button for this. Maybe there is also one used in the dashboard already?
And always: we do not use colours directly in the components, but use variables.

fuzzylogic2000 (commented):
Perfect! 
Apart from the styling, which is only nearly perfect. :) Will fix that.

adhocracy-plus/assets/scss/components/_userdashboard.scss
fuzzylogic2000: Ok, I get why you added that now! Because the buttons we have a differently coloured border. :grimacing: 
But as far as I can remember, we are meant to use the same button as in the dashboard for the manual. I changed it to that now, will merge and then add an issue to discuss with @mcastro-lqd what to do about that!

PR #292, opened 2022-08-05
Ks 2022 08 sample project

Rineee:
This is branched off of the branch here: https://github.com/liqd/a4-kosmo/pull/291 as they both added migrations to cms/settings. So only the last 2 commits are relevant here.. hope that made sense to do it this way..


fuzzylogic2000 (approved):
Totally works! Nice!
Does someone else want to check? I will assign others to review!

apps/users/signals.py
```diff
@@ -0,0 +1,19 @@
+from allauth.account.signals import email_confirmed
+from django.dispatch import receiver
+
+from adhocracy4.follows.models import Follow
+from adhocracy4.projects.models import Project
+from apps.cms.settings.models import OrganisationSettings
+
+
+@receiver(email_confirmed)
+def auto_follow_sample_project(request, email_address, **kwargs):
```
goapunk, line 10: I'm unsure about the best approach here. Using `email_confirmed` here has some implications: it will trigger every time  the user changes their email (which currently will throw an `IntegrityError` -> `UNIQUE constraint failed: a4follows_follow.project_id, a4follows_follow.creator_id`.
If the user is still following the project that's fine, if not they shouldn't be auto-following again. I guess we can add a check here if the user is already following and do nothing if so. Or use a signal which is only triggered on account creation.

goapunk (changes requested):

apps/users/signals.py
fuzzylogic2000: True! I will do the welcome email now, and think, we can probably use the same signal for triggering this and the welcome email. So, maybe I just go on with that in this branch!

fuzzylogic2000: No, right, we are doing the same in mB and should also fix it there. https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/users/signals.py

fuzzylogic2000: Well, actually, I think, we need both:
1. Here we should use a signal that is sent on user create as it doesn't matter if the user never logs in, but autofollows a project. Or we don't use a signal and just do it directly in the save method of the signup form.
2. For the welcome email we should check if that is the first email address for the user.
I will start with 2. and the email now.

fuzzylogic2000:
@Rineee I used your code for creating the Follow, but moved it from the signal to the registration, when the user is created.

fuzzylogic2000 (approved):
Yes, I am also very fine with that!

PR #294, opened 2022-08-08
styling/user profile and dashboard: use same dark border colour for u…

### LINKED ISSUE #293: user dashboard's manual button does not fit the profile's box above
**URL:** 
user's dashboard: https://kosmo-dev.liqd.net/userdashboard/overview/
project dashboard: https://kosmo-dev.liqd.net/liqd/dashboard/organisations/projects/
profile: https://kosmo-dev.liqd.net/profile/katharina/
**user:** 
**expected behaviour:** I think, we wanted to take the same button for the manual link in the user's dashboard as in the project dasboard. When we do that, the borders around the button and the user info box don't have the same colour. (see screenshot)
**behaviour:** The border for the user's info box is from styling from the profile view. So, the question is: which colour? Can we change the border-colour in the profile as well, so that it is the same everywhere? Because if we change it for the manual's button, it will be different from the left nav in the project dashboard.
**important screensize:**
**device & browser:** 
**Comment/Question:** Hope, that does make sense!

user's dashboard: 
Screenshot from 2022-08-08 15-20-24

project dashboard: 
Screenshot from 2022-08-08 15-20-49

profile: 
Screenshot from 2022-08-08 15-22-21

@mcastro-lqd Please help and tell us what to do! (Very short form: can we also make the border in the user profile darker?)

fuzzylogic2000:
…ser info box as in btn--light - fixes #293

### DIFF
--- a/adhocracy-plus/assets/scss/components/_userdashboard.scss
+++ b/adhocracy-plus/assets/scss/components/_userdashboard.scss
@@ -4,7 +4,7 @@
 }
 
 .userdashboard-profile__user {
-    border: 1px solid #dadada;
+    border: 1px solid $border-color;
     text-align: center;
     padding-bottom: 2 * $padding;
 }


phillimorland (approved):
nice!

PR #296, opened 2022-08-09
[6271] Kt 2022 08 latest project comments

khamui:
- [x] plural translation ( 1 new comment, 0 .. n new comments)
- [x] slick showing more tiles

### DIFF
--- a/adhocracy-plus/assets/js/app.js
+++ b/adhocracy-plus/assets/js/app.js
@@ -67,7 +67,7 @@ function init () {
       {
         breakpoint: 1024,
         settings: {
-          slidesToShow: 2,
+          slidesToShow: 3,
           arrows: true
         }
       }
--- a/adhocracy-plus/assets/scss/_layout.scss
+++ b/adhocracy-plus/assets/scss/_layout.scss
@@ -9,7 +9,7 @@
     @supports (display: grid) {
         display: grid;
         grid-column-gap: _grid-get("gutter", $settings);
-        grid-template-columns: repeat($n, 1fr);
+        grid-template-columns: repeat($n, minmax(0, 1fr));
 
         > * {
             margin-left: 0 !important;
--- a/adhocracy-plus/assets/scss/components/_project_tile_carousel.scss
+++ b/adhocracy-plus/assets/scss/components/_project_tile_carousel.scss
@@ -4,7 +4,7 @@
 
     // overwrites for tiles in carousel
     .tile.organisation__tile {
-        height: 430px;
+        height: 440px;
         margin: 0;
         display: flex !important;
     }
@@ -24,6 +24,7 @@
 
     .slick-track {
         padding-bottom: $padding;
+        margin-left: 0;
     }
 
     .slick-prev,
@@ -68,10 +69,6 @@
 @media screen and (min-width: $breakpoint-md) {
     .project-tile-carousel {
         margin: $spacer 0;
-
-        .tile.organisation__tile {
-            height: 570px;
-        }
     }
 
     .project-tile-carousel__wrapper {
--- a/apps/projects/helpers.py
+++ b/apps/projects/helpers.py
@@ -1,4 +1,7 @@
+from datetime import timedelta
+
 from django.db.models import Q
+from django.utils import timezone
 
 from adhocracy4.comments.models import Comment
 from apps.classifications.models import AIClassification
@@ -46,3 +49,9 @@ def get_num_ai_classifications(project):
 def get_num_classifications(project):
     return get_num_user_classifications(project) + \
         get_num_ai_classifications(project)
+
+
+def get_num_latest_comments(project, until={'days': 7}):
+    all_comments_project = get_all_comments_project(project)
+    return all_comments_project.filter(created__gte=timezone.now() -
+                                       timedelta(**until)).count()
--- a/apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
+++ b/apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
@@ -21,8 +21,7 @@
             <div class="tile__private-badge">
                 <div class="tile__private-badge--top"></div>
                     <i class="fas fa-eye" title="{% trans 'Project is semi-public.' %}" aria-label="{% trans 'Project is semi-public.' %}"></i>
-                <div class="tile__private-badge--left"></div>
-            </div>
+                <div class="tile__private-badge--left"></div> </div>
 
           {% endif %}
         </div>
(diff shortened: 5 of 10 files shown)


phillimorland:
@khamui shall I review this or are you finishing checklist?

khamui:
> @khamui shall I review this or are you finishing checklist? also final task should be able to be completed with changes `slidesToShow: 2,` in app.js and maybe changing styling width

sorry forgot to set WIP, i am about to push the slick stuff...will make a separate commit, because not sure if that is how it is meant to be what i am doing there

adhocracy-plus/assets/scss/components/_project_tile_carousel.scss
```diff
@@ -24,6 +24,7 @@
 
     .slick-track {
         padding-bottom: $padding;
+        margin-left: 0;
```
khamui, line 27: that makes it align left (only affects if we have less projects than 3), otherwise it looks a bit off

khamui:
@phillimorland okay ready to be reviewed..

apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
```diff
@@ -31,27 +31,30 @@
...
+                    {% plural %}
+                        new comments
+                    {% endblocktrans %}
+                </span>
             </div>
-            <div class="d-flex justify-content-between">
-                <span class="status-bar__status"><i class="fas fa-clock" aria-hidden="true"></i>
+            {% endif %}
+            <div class="d-flex justify-content-between my-1">
+                <span class="status-bar__status"><i class="far fa-clock" aria-hidden="true"></i>
```
phillimorland, line 51: when there is a longer title everything is pushed out of the tile, on design tiles are different ratio so appear taller, so we can maybe do same here?

adhocracy-plus/assets/scss/components/_project_tile_carousel.scss
```diff
@@ -68,10 +69,6 @@
 @media screen and (min-width: $breakpoint-md) {
     .project-tile-carousel {
         margin: $spacer 0;
-
-        .tile.organisation__tile {
-            height: 570px;
-        }
```
phillimorland, line 74: this might be why they look weird on organisation page now?

apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
```diff
@@ -28,9 +28,9 @@
         </div>
     </div>
     <div class="tile__body tile__body-status">
-        <h3 class="tile__title" id="project-title-{{ project.pk }}">
+        <h4 class="tile__title" id="project-title-{{ project.pk }}">
             <a href="{{ project|project_url }}">{{ project.name }}</a>
-        </h3>
+        </h4>
```
phillimorland, line 33: for a11y we shouldn't actually change the tag as it is the third heading on the page, we can change the size with css though

adhocracy-plus/assets/js/app.js
```diff
@@ -62,12 +62,12 @@ function init () {
     mobileFirst: true,
     infinite: false,
     variableWidth: false,
-    slidesToShow: 1.2,
+    slidesToShow: 2,
```
phillimorland, line 65: this doesn't work on small screens?!

phillimorland (changes requested):
Cool, few changes, also wondering should we also change the tiles on `/userdashboard/overview/following/` as it's weird that they are different when you just click 'view more' think it just needs `l-tiles-3` or something in template.

apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
```diff
@@ -31,27 +31,30 @@
         <h3 class="tile__title" id="project-title-{{ project.pk }}">
             <a href="{{ project|project_url }}">{{ project.name }}</a>
         </h3>
-        <p class="tile__text">{{ project.description }}</p>
```
phillimorland, line 34: This tiles are used in the organisation page as well so we still need to show the description in that case, I think you can can use something like that?! 
```
{% if request.path != '/userdashboard/' %}
```

apps/userdashboard/views.py
```diff
@@ -55,9 +59,21 @@ def actions(self):
...
+            project.latest_comments_num = self._latest_comments_num(
+                project)
+            projects.append(project)
+
+        return projects
+
+    def _latest_comments_num(self, project):
+        all_comments_project = helpers.get_all_comments_project(project)
+        return all_comments_project.filter(created__gte=datetime.now() -
+                                           timedelta(days=7)).count()
```
phillimorland, line 76: do we want to show the most comments or the most recent? it's currently serving the project with the most comments first, not the project with most recent comments, don't know if we care?!

apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
khamui: yes i checked it and the title of this section is already `h3`, so i thought it would make sense as these are nested. But i don't know if `h2` or `h1` are used though

phillimorland: @khamui it looks like we messed up with that, I think we should have 'following' as h2 and maybe have a hidden h1 saying 'user overview' or something but maybe for now just sort out the h2 and h3?

apps/userdashboard/views.py
```diff
@@ -55,9 +59,21 @@ def actions(self):
     def projects_carousel(self):
         sorted_active_projects, sorted_future_projects, sorted_past_projects =\
             self.request.user.get_projects_follow_list()
-        return (list(sorted_active_projects) +
-                list(sorted_future_projects) +
-                list(sorted_past_projects))[:8]
+        display_projects = (list(sorted_active_projects) +
+                            list(sorted_future_projects) +
+                            list(sorted_past_projects))[:8]
+        projects = []
```
Rineee, line 65: I think you dont need this new list here, you can just go through the old list and add the property..

adhocracy-plus/assets/scss/components/_project_tile_carousel.scss
khamui: just checked, it is inside of `.project-tile-carousel` so no effect on organisation

adhocracy-plus/assets/scss/_layout.scss
```diff
@@ -9,7 +9,7 @@
     @supports (display: grid) {
         display: grid;
         grid-column-gap: _grid-get("gutter", $settings);
-        grid-template-columns: repeat($n, 1fr);
+        grid-template-columns: repeat($n, minmax(0, 1fr));
```
khamui, line 12: this is a very fundamental change to the grid. I had the issue described in the following link, which made one column wider than the other two.

Here the excerpt that is relevant:

>The common answer of repeat(3, 1fr) is not quite correct.
>
>This is because 1fr is about the distribution of available(!) space. This breaks as soon as the content becomes >bigger than the track size. By default, it does not overflow and adjust the column width accordingly. That's why >not all 1fr are guaranteed to be of equal width. 1fr is actually rather just a shorthand for minmax(auto, 1fr).
>
>If you really need the columns to be the exact same width you should use:

>` grid-template-columns: repeat(3, minmax(0, 1fr));`

>minmax(0, 1fr) allows the grid tracks to be as small as 0 but as large as 1fr, creating columns that will stay >equal. But, be aware that this will cause overflows if the content is bigger than the column or cannot be >wrapped.

The overflows should be handled? what do you think @phillimorland ?

link to [stackoverflow answer](https://stackoverflow.com/questions/47601564/equal-width-columns-in-css-grid):

apps/projects/helpers.py
```diff
@@ -46,3 +49,9 @@ def get_num_ai_classifications(project):
 def get_num_classifications(project):
     return get_num_user_classifications(project) + \
         get_num_ai_classifications(project)
+
+
+def get_num_latest_comments(project, until={'days': 7}):
+    all_comments_project = get_all_comments_project(project)
+    return all_comments_project.filter(created__gte=timezone.now() -
+                                       timedelta(**until)).count()
```
Rineee, line 57: Cool, much better place for that logic!

apps/userdashboard/views.py
```diff
@@ -35,8 +36,13 @@ def organisations(self):
...
     def projects(self):
-        return Project.objects.filter(follow__creator=self.request.user,
-                                      follow__enabled=True)
+        projects = Project.objects.filter(follow__creator=self.request.user,
+                                          follow__enabled=True)
+        for project in projects:
+            project.latest_comments_num =\
+                helpers.get_num_latest_comments(project)
+
+        return projects
```
Rineee, line 45: Nice, thanks!

apps/userdashboard/views.py
```diff
@@ -39,8 +36,13 @@ def organisations(self):
...
```
Rineee, line 45: Nice, thanks!

Rineee:
Cool, Im happy with the python part now, thanks! But you could remove the first commit now? Or squash parts of the third with it?

khamui:
@phillimorland progressbar is restored and only showing if not on overview page or following and image ratio is identical as in carousel.

@Rineee fixup commits and rebased 🤝

Rineee (approved):
Works great! But now its all in one commit :see_no_evil: If we leave it like this, maybe at least change commit message? Sorry for nit-picking!

phillimorland:
@khamui @Rineee is the commit message now fixed?

khamui:
> @khamui @Rineee is the commit message now fixed?

yes i changed it

phillimorland (approved):
cool!

PR #297, opened 2022-08-09
[6268] apps/user: send send in blue welcome email after email confirm

fuzzylogic2000:
When this is finished, I will fix the management command to send all emails again, but for now and for testing this is easier, because to test:
- add the api key to your local settings
- and use the management comand `send_test_emails`

`python manage.py send_test_emails <your_email>`

~~The email templates are a bit broken, though, and the link and actual username are only added to the English template. And there is a double greeting in the English one, because I could not get the `default` working with the params and for the contact I could only fill in the email on the api-call side.~~

To do:

- [x] only send to users confirming their first email
- [x] fix management command after reviewing
- [x] fix tests

### DIFF
--- a/apps/users/emails.py
+++ b/apps/users/emails.py
@@ -1,12 +1,18 @@
 from email.mime.image import MIMEImage
 
+import sib_api_v3_sdk
 from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpResponse
 from django.template.loader import get_template
 from django.urls import reverse
 from django.utils import translation
 from django.utils.translation import gettext_lazy as _
+from sentry_sdk import capture_exception
+from sib_api_v3_sdk.rest import ApiException
 
 from adhocracy4.emails import Email
+from adhocracy4.emails.mixins import SyncEmailMixin
 
 from .models import User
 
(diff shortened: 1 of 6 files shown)


apps/users/emails.py
```diff
@@ -85,3 +91,67 @@ def get_html_link(self, link_text, url):
...
+
+    def get_overview_link(self):
+        url = reverse('userdashboard-overview')
+        link = self.get_host() + url
+        return link
+
+    def get_template_id(self, receiver):
+        language = self.get_receiver_language(receiver)
+        try:
+            template_id = int(settings.SENDINBLUE_TEMPLATES[language])
```
goapunk, line 110: why do we need the int here? and could it only be done at the return instead of in both conditions?

fuzzylogic2000: We shouldn't need it, but I didn't want to rely too much on the settings. On the other hand, if that goes wrong, we will notice it and the int only works if someone adds the numbers as strings, so will just remove it and hope for the best. :p

tests/users/test_views.py
```diff
@@ -57,6 +58,9 @@ def test_logout_with_next(user, client, logout_url):
 
 
 @pytest.mark.django_db
+@override_settings(
+    SENDINBLUE_API_KEY=None
+)
```
fuzzylogic2000, line 63: Fixing the test, it's easy to see, that we should probably add sth. to test that the settings are also set properly, not only if they are set. But I think, we can do that in a second iteration together with adding tests.

fuzzylogic2000:
I think this is good to go now, and we can add tests and improve the error handling in a new PR.

phillimorland (commented):
not working, receiving all except welcome email but assume that's due to `management command` being fixed, you guys can figure out merge ect.

fuzzylogic2000:
> not working, receiving all except welcome email but assume that's due to `management command` being fixed, you guys can figure out merge ect.

It should still work, but the failing is probably due to the silently failing, when sth. is wrong with the token (same problem as in tests). I will investigate now!
And I will also add sth. to the test email-thing, printing that it actually attempts to send the email, so that we know if the problem is there or in the actual API call.

fuzzylogic2000:
@phillimorland Could you try again? When you send the test emails in the terminal it should say at the end of all the messages
```Sending send in blue welcome email
{'message_id': '<202208100840.90562644453@smtp-relay.mailin.fr>',
 'message_ids': None}```
when it's fine and 
```Sending send in blue welcome email
Exception when calling TransactionalEmailsApi->send_smtp_email: (401)
Reason: Unauthorized
HTTP response headers: HTTPHeaderDict({'Date': 'Wed, 10 Aug 2022 06:42:30 GMT', 'Content-Type': 'application/json', 'Content-Length': '50', 'Connection': 'keep-alive', 'sib-request-id': '9d478ae8-8790-9fdd-b78f-068422b592cb', 'access-control-allow-origin': '*', 'access-control-allow-credentials': 'true', 'access-control-allow-headers': '*', 'access-control-allow-methods': 'PUT, POST, GET, DELETE, PATCH, OPTIONS', 'x-envoy-upstream-service-time': '1', 'x-sib-server': 'srv-pr-online-rancher-worker-5.onl.51b.tech', 'x-content-type-options': 'nosniff', 'x-xss-protection': '1', 'CF-Cache-Status': 'DYNAMIC', 'Expect-CT': 'max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"', 'Server': 'cloudflare', 'CF-RAY': '7386c07e5f17aca7-TXL'})
HTTP response body: {"message":"Key not found","code":"unauthorized"}```
when it's not. Then sth. is probably wrong with the API key.

@goapunk and that is also why the tests work: it does go into the except block, but the tests don't care as we do not raise an exception, but capture it. So, should I change that?

**Comment by phillimorland** [comment]
@fuzzylogic2000 yep now receiving 
```
Sending send in blue welcome email
{'message_id': '<202208100928.23701099785@smtp-relay.mailin.fr>',
 'message_ids': None}
```

File: tests/users/test_views.py [hunk 3]
```diff
@@ -57,7 +58,8 @@ def test_logout_with_next(user, client, logout_url):
 
 
 @pytest.mark.django_db
-def test_register(client, signup_url):
+@patch('apps.users.welcome.WelcomeEmail.dispatch', return_value='/')
```
fuzzylogic2000, line 61: I found a better way to fix the tests and not send a wrong key to the API with every test call.

goapunk: nice!

goapunk (approved):

PR #303, opened 2022-08-11
[#6272] reaction feed frontend

phillimorland:


apps/actions/templates/a4_candy_actions/includes/reaction.html
```diff
@@ -0,0 +1,29 @@
+{% load i18n contrib_tags %}
+{% if action.type == 'comment' and action.verb == 'add' %}
```
phillimorland, line 2: so i did this because otherwise it was showing when i added a project when i had following at the end of the `if inside the action__line and when I removed it, it had an empty space, not sure whats best to do?!
```
{% else %}
  {{ action }}
{% endif %}
```

Rineee:
I cherry picked the commits to my PR https://github.com/liqd/a4-kosmo/pull/300 as couldnt make it work otherwise :see_no_evil:

PR #306, opened 2022-08-11
ModerationNotificationList: remove Alert, adding to ModerationNotific…

### LINKED ISSUE #161: Moderation Dashboard – Warning Message
**URL:** https://kosmo-dev.liqd.net/userdashboard/moderation/detail/rines-testproject/
**important screensize:** *Mobile & Desktop*
**device & browser:** *Safari Version 15.0 (16612.1.29.41.4, 16612)*
**Comment/Question:** *Is it easy to put the warning message where the tile is? Cause currently if we have a long tile list, the message is displayed on top and the user can't see it if you don't scroll back to the top. I think it would make more sense if its displayed where the action took place, right?* 

Currently: 
Screenshot 2021-11-16 at 11 18 26

How it could be:
Screenshot 2021-11-16 at 11 18 47

khamui:
…ation

closes #161

### DIFF
--- a/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
+++ b/apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
@@ -6,6 +6,8 @@ import { ModerationStatement } from './ModerationStatement'
 import { ModerationNotificationActionsBar } from './ModerationNotificationActionsBar'
 import { alert as Alert } from 'adhocracy4'
 
+const alertTime = 6000
+
 const translated = {
   statementAdded: django.pgettext('kosmo', 'Your feedback was successfully delivered.'),
   statementEdited: django.pgettext('kosmo', 'Your feedback was successfully updated.'),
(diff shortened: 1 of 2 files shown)


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -181,9 +197,17 @@ export const ModerationNotification = (props) => {
...
+        type: 'error',
+        message: error,
+        timeInMs: 3000
+      })
     } else {
-      props.onChangeStatus(alertMessage)
+      setAlert({
+        type: 'success',
+        message: alertMessage,
+        timeInMs: 3000
```
phillimorland, line 209: I#m a little torn about the timeout since we were told it isn't very accessible, maybe you can make it a bit longer?

phillimorland (commented):
@khamui I think maybe increase the times as that's very quick and maybe confuse older users, and sort the merge conflict then good to go

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
khamui: i think if we do not specify timeInMs it will stay there until user dismisses it. think that is the best way in terms of a11y? @phillimorland otherwise i would increase to 6 sec.?

khamui: i went with the 6 secs. for now :)

phillimorland:
coolio!

PR #309, opened 2022-08-15
Ks 2022 08 mod statement action

Rineee:
Here are some todos, but for the blocked comments I would say merge first and then see what PMs say when testing

- [ ] check icon and maybe replace
- [ ] ask PM about blocked comments

### DIFF
--- a/apps/userdashboard/views.py
+++ b/apps/userdashboard/views.py
@@ -10,6 +10,7 @@ from adhocracy4.polls.models import Poll
 from adhocracy4.projects.models import Project
 from adhocracy4.rules import mixins as rules_mixins
 from apps.documents.models import Chapter
+from apps.moderatorfeedback.models import ModeratorCommentStatement
 from apps.organisations.models import Organisation
 from apps.users.models import User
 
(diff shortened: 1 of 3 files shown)


apps/userdashboard/views.py
```diff
@@ -65,10 +67,19 @@ def actions(self):
...
+                                    and action.actor != user]
+        statement_actions = Action.objects.filter(
+            obj_content_type=ContentType.objects.get_for_model(
+                ModeratorCommentStatement),
+        )
+        filtered_statement_actions = [action for action in statement_actions if
+                                      action.obj.comment.creator == user
+                                      and action.actor != user]
+        return sorted(filtered_comment_actions + filtered_statement_actions,
+                      key=lambda action: action.timestamp, reverse=True)
```
Rineee, line 82: This is somewhat complicated, but did not know how else to do this as the actions have different properties depending on the obj_content_type..

phillimorland (approved):
works like a charm and I think views.py seems fine for me as definitions are clear

Issue #322, opened 2022-08-19
Welcome Mail, button to getting started with wrong link


rilanari:
**URL:** Welcome Mail in Sendinblue
**user:** Participant
**expected behaviour:** The 'Get Started now' button directs you to the Manuals wegtail page
**behaviour:** The 'Get Started now' button directs you to My Overview
**important screensize:** Mail
**device & browser:** All 
**Comment/Question:** Please, make the button go to the Manuals (which now has the terms of use)

Screenshot?

rilanari:
I solved it myself! I hope it works!!! XD

Rineee:
@rilanari Did you test it? I think you would need to get the link from the backend, so we would have to provide it for you..

rilanari:
yes! it does work!!! :)

Rineee:
> yes! it does work!!! :)

But what did you link it to? Because the manual page is settable, so if someone changes the link there, it would not work anymore right?

rilanari:
Oh I linked it to this: https://kosmo-dev.liqd.net/use-case-test-organisation/nutzungsbedingungen/

rilanari:
i guess it will change?

rilanari:
It is the same link from the button from "Getting Started" in "My Overview"

PR #323, opened 2022-08-22
users/email: add manual page to welcome email context

Rineee:
see https://github.com/liqd/a4-kosmo/issues/322

I also already adapted the templates.. I tried to add a if else statement there, to link to the overview page is manual page is not set, but that did not work..

### DIFF
--- a/apps/users/emails.py
+++ b/apps/users/emails.py
@@ -10,9 +10,11 @@ from django.utils import translation
 from django.utils.translation import gettext_lazy as _
 from sentry_sdk import capture_exception
 from sib_api_v3_sdk.rest import ApiException
+from wagtail.models import Site
 
 from adhocracy4.emails import Email
 from adhocracy4.emails.mixins import SyncEmailMixin
+from apps.cms.settings.models import ImportantPages
 
 from .models import User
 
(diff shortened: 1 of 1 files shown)


apps/users/emails.py
```diff
@@ -104,6 +106,17 @@ def get_overview_link(self):
...
+    def get_manual_page_link(self):
+        site = Site.objects.filter(
+            is_default_site=True
+        ).first()
+        important_pages = ImportantPages.for_site(site)
+        if getattr(important_pages, 'platform_information') and \
+                getattr(important_pages, 'platform_information').live:
+            url = getattr(important_pages, 'platform_information').url
+            return self.get_host() + url
+        return None
```
fuzzylogic2000, line 118: If using if/else in the template does not work, maybe we can return another link here? Like the landing page?
(And also if the if/else does not work, we cannot really use sendinblue for all emails, right? But that's another discussion and another point in time.)

fuzzylogic2000 (approved):
Looks good!

apps/users/emails.py
Rineee: Yes, totally, we should return sth else then! Maybe the overview page? But let me try again with if/else in template first, I also was a bit impatient on Monday and should probably give it another go..

Rineee: @fuzzylogic2000 I still could not make if/else in template work :woman_shrugging: So, it now returns the overview link if manual page is not set.. what do you think?

goapunk: doing it this way has the advantage that it's all in one place I'd think, and not somewhere buried in a template at sendinblue which I will totally have forgotten and then will search forever :sweat_smile:

goapunk:
will leave for @fuzzylogic2000 to merge as still one open question

apps/users/emails.py
fuzzylogic2000: Yeah, but we would like to not display the whole sentence when we don't have the link. But we also don't want to hand over the sentence as a variable. So, this with the fallback link is the best solution, I think!

fuzzylogic2000 (approved):
Nice!

PR #324, opened 2022-08-29
Ks 2022 08 translations

Rineee:
I left out the translation for 'Getting Started' as I didnt really know how to translate it and did not get an answer from PMs, so we might have to add that later.. sorry!


fuzzylogic2000 (approved):
Looks good!

